PR #9, opened 2016-01-27
View by tag

shiboying:


app/assets/javascripts/tags.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
```
rmacklin: I know this file came from the rails generator, but I would suggest removing it. We may add some javascript for tags, but that will come later (and it won't be coffeescript), so my preference would be to only commit files we're currently using.

app/assets/stylesheets/tags.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the Tags controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
```
rmacklin: Same as above: I wouldn't commit this generated file if we aren't using it yet.

app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def index
+    @tags = ActsAsTaggableOn::Tag.all.map(&:name)
+  end
+
+  def show
+    picture_ids = ActsAsTaggableOn::Tagging
+      .joins("join tags on tags.id = taggings.tag_id and tags.name = '#{@tag}'")
+      .where(taggable_type: 'Picture').map(&:taggable_id)
+    @pictures = Picture.find(picture_ids)
```
rmacklin: Are these lines equivalent to `Picture.tagged_with(@tag)`?

app/helpers/tags_helper.rb
```diff
@@ -0,0 +1,2 @@
+module TagsHelper
+end
```
rmacklin: Same as above, I wouldn't commit this generated file if we aren't using it yet.

app/views/layouts/_footer.html.erb
```diff
@@ -0,0 +1,14 @@
...
+        <h1 class="footer__heading">About</h1>
+          <p class="m-b-0">Content for about</p>
+      </div>
+      <div class="col-sm-6">
+        <h1 class="footer__heading">Contact</h1>
+        <p class="m-b-0">Content for contact</p>
+      </div>
+    </div>
+  </div>
+</footer>
```
rmacklin: Along the lines of atomic commits as mentioned in https://github.com/AppFolioOnboarding/Boying_ImageUploader/commit/e3fbf379d903525743de0061bd9be450138513df#commitcomment-15696228, it's also recommend to create atomic branches. In this case, I would probably opt for making the bootstrap and style changes on a separate branch from the one that adds the tags views, because these are distinctly separate pieces (and thus it makes sense to review them separately). I just realized there isn't a user story for adding the styles, so we should add one, and that way our PRs/branches will still map to individual stories.

app/views/tags/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+        <%= link_to (image_tag picture.link, class: 'card-img-top'), picture %>
+        <div class="card-block">
+          <% picture.tag_list.each do |tag| %>
+            <%= link_to tag, tag_path(tag) %>
+          <% end %>
+        </div>
+      </div>
+    </div>
+  <% end %>
+</ul>
```
rmacklin: This markup looks [veryyyy familiar](https://github.com/AppFolioOnboarding/Boying_ImageUploader/blob/e3fbf379d903525743de0061bd9be450138513df/app/views/pictures/index.html.erb#L7-L20) :wink:. We should always aim to keep our code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). We could achieve that goal pretty easily by extracting a partial for the shared markup.

However, (we should discuss this tomorrow, but) I don't think this is a proper use of the `show` action. Currently, `tags#show` is effectively the same view as `pictures#index` (just filtered by the selected tag). It would be better if that was implemented in `pictures#index` via query parameters (e.g. `/pictures?tag=banana`). Then, `tags#show` would be changed from _“show me the pictures associated with this tag"_ to _"show me this tag"_. This page could have edit and delete links for the tag resource. It could include a link to `pictures#index` filtered by that tag, even including a preview of a few of the images with that tag, but I don't think it should list all the associated images, which is the responsibility of `pictures#index`.

app/assets/javascripts/tags.coffee
shiboying: Good point, I remove them in a new commit.

app/views/layouts/_footer.html.erb
shiboying: I have removed all UI related things in this branch and committed them in a new branch which I have not yet pushed. And I separated the last commit into several ones. Thanks for pointing it out!

app/views/tags/show.html.erb
shiboying: Yes, good catch. I made a partial image list for them.

app/controllers/tags_controller.rb
shiboying: Yes, I changed it to this.

app/assets/javascripts/tags.coffee
rmacklin: Thanks!

app/views/layouts/_footer.html.erb
rmacklin: Excellent, thanks for separating that.

app/views/tags/show.html.erb
rmacklin: So, that's definitely an improvement to DRY the code. But, I'm more concerned with the fact that we're misusing `tags#show` by making it duplicate the functionality of `pictures#index`. What are your thoughts on this:

> However, (we should discuss this tomorrow, but) I don't think this is a proper use of the `tags#show` action. Currently, `tags#show` is effectively the same view as `pictures#index` (just filtered by the selected tag). It would be better if that was implemented in `pictures#index` via query parameters (e.g. `/pictures?tag=banana`). Then, `tags#show` would be changed from _“show me the pictures associated with this tag"_ to _"show me this tag"_. This page could have edit and delete links for the tag resource. It could include a link to `pictures#index` filtered by that tag (even going as far as including a preview of a few of the images with that tag), but I don't think it should list all the associated images, which is the responsibility of `pictures#index`.

app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def show
+    @pictures = Picture.tagged_with(@tag)
+  end
+
+  private
+
+  def find_tag
+    @tag = ActsAsTaggableOn::Tag.where(name: params[:id]).first
```
rmacklin: It's a bit confusing that we call the param `id` even though it's not a `Tag` `id` but rather it's a `Tag` `name`. We can [override the named route parameter](http://guides.rubyonrails.org/routing.html#overriding-named-route-parameters) to make this clearer (i.e. to make it `Tag.where(name: params[:name])` instead of `Tag.where(name: params[:id])`).

app/views/pictures/index.html.erb
```diff
@@ -2,16 +2,6 @@
...
-    <li>
-      <%= link_to (image_tag picture.link, class: 'image'), picture %>
-      <br>
-      <% picture.tag_list.each do |tag| %>
-        <%= tag %><br>
-      <% end %>
-    </li>
-  <% end %>
-</ul>
+<%= render partial: 'list/image_list', object: @pictures %>
```
rmacklin: When you pass something as `object`, it is available in the partial as a local variable matching the name of the partial. So in this case, it's exposed in the partial as `image_list`. However, you're not using that in the partial (you're referring directly to the `@pictures` instance variable), so that's a bit confusing.

However, I don't actually think we should be extracting this partial right now. Refer to https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/9#discussion_r51034242

app/views/tags/show.html.erb
shiboying: I agree that `tags#show` should have some functionalities about edit and delete links for this tag. But I think it makes more sense to me that this page should have pictures associated with this tag shown. If you want to edit links for the tag without showing those pictures(for preview), it is not convenient I think. So assume we do need `tags#show` to display pictures(or just preview for those links), why not just use this page for showing all pictures with this tag?

app/views/tags/index.html.erb
```diff
@@ -0,0 +1,7 @@
+<ul id="tags_list">
+  <% @tags.each do |tag| %>
+    <li>
+      <%= link_to tag, tag_path(tag) %>
```
rmacklin: This isn't a big deal but I just thought I'd mention it:

If you removed the `.map(&:name)` from [this line](https://github.com/AppFolioOnboarding/Boying_ImageUploader/blob/1d748d01f01d22b0502b8fa4bd3b93796bb20901/app/controllers/tags_controller.rb#L5), then you could rewrite this as:
```ruby
link_to tag.name, tag
```
which may be a bit cleaner.

shiboying: I think I override the named route parameter but this doesn't work.
However, this `<%= link_to tag, tag_path(tag.name) %>` works.

Can I use this?

app/controllers/tags_controller.rb
shiboying: Yes, I agree it is a little confusing. I will change it.

app/views/pictures/index.html.erb
shiboying: Yes, thanks for the explanation. That's a good recall for me. Will fix. And I wrote my thought under that comment, please check.

app/views/tags/show.html.erb
rmacklin: I agree that it is useful to have a page that shows the pictures associated with a tag. My argument is that this page should be `pictures#index` with a query parameter (e.g. `/pictures?tag=banana`). The responsibility of the `pictures#index` action (and generally any resource's index action) is to _"list all pictures"_. When you add a query string, it becomes _"list all pictures filtered by this query"_, but conceptually the URL `/pictures` is the sole owner of the _"list pictures"_ responsibility of our application (rather than spreading this responsibility across multiple URLs).

By not spreading the _"list pictures"_ responsibility over multiple actions, we increase the maintainability of our application, which is great for us as developers. Additionally, I think it makes our application better follow semantic URLs, which is great for _all consumers_ of the application (including end-users and even robots).

@dtognazzini Maybe you have some other comments since you were the one to suggest this?

dtognazzini: Thinking in terms of REST, both a "picture" and  "a tag" are resources. The CRUD operations on any resource have predefined functionality. Maintaining the expected functionality for RESTful operations is crucial for REST as a whole to work.

Similar to the CRUD operations, all of the Rails actions (including "show" and "index") also have predefined or expected functionality. In the case of "show", the expected functionality is to implement the R, or "read"  operation from CRUD. For "index", the expected functionality is to implement an interface to query the collection of resources.

Why does this matter?

In the future, you may want to interact with a given resource using a different data format. Perhaps you want to retrieve information about a resource, but for whatever reason, you need the data back in JSON. This is where REST shines. You should be able to reuse the existing URL for reading (the "show" action in Rails) and specify a content type according to your needs, JSON in this case.

app/views/tags/index.html.erb
rmacklin: We discussed this is person, but I'm just noting it here: yes, go ahead and use `link_to tag.name, tag_path(tag.name)` because otherwise we'd have to monkey patch `ActsAsTaggable::Tag` to override `model_name` and `to_param` in order to make the "Rails magic" work.

app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def show
+    @pictures = Picture.tagged_with(@tag)
+  end
+
+  private
+
+  def find_tag
+    @tag = ActsAsTaggableOn::Tag.where(name: params[:name])
```
rmacklin: If you use `find_by` instead of `where` here then you won't need to call `.first` [here](https://github.com/AppFolioOnboarding/Boying_ImageUploader/blob/5199d15112acca0eb0a30844fb54c653d281050e/app/views/tags/show.html.erb#L1)

config/routes.rb
```diff
@@ -5,6 +5,7 @@
   # You can have the root of your site routed with "root"
   root 'pictures#index'
   resources :pictures, only: [:new, :create, :index, :show, :destroy, :edit, :update]
+  resources :tags, as: :acts_as_taggable_on_tag, only: [:index, :show], param: :name
```
rmacklin: Personally I think it's cleaner without `as: :acts_as_taggable_on_tag` since then you could just use `tag_path` instead of `acts_as_taggable_on_tag_path` [here](https://github.com/AppFolioOnboarding/Boying_ImageUploader/blob/5199d15112acca0eb0a30844fb54c653d281050e/app/views/list/_image_list.html.erb#L7).

app/views/tags/show.html.erb
shiboying: Got it. And I will take it into consideration in future work. For this one, I will make a new branch to modify the routes. Thanks guys!

rmacklin: Changes made in #10

PR #10, opened 2016-01-28
View pictures by tag with pictures controller index action

shiboying:


app/controllers/pictures_controller.rb
```diff
@@ -1,8 +1,9 @@
 class PicturesController < ApplicationController
-  before_filter :find_picture, :only => [:update, :show, :edit, :destroy]
+  before_action :find_picture, :only => [:update, :show, :edit, :destroy]
 
   def index
-    @pictures = Picture.all.order('updated_at DESC')
+    @params_tag = params[:tag]
+    @params_tag ? @pictures = Picture.tagged_with(params[:tag]) : @pictures = Picture.all.order('updated_at DESC')
```
bboe: How do you feel about this statement? Is it easy to read?

Either way, consider:

    @picutres = @params_tag ? Picture.tagged_with(params[:tag]) : Picture.all.order('updated_at DESC')

Also does `tagged_with` provide the same ordering?

app/controllers/tags_controller.rb
```diff
@@ -1,13 +1,9 @@
...
-  def show
-    @pictures = Picture.tagged_with(@tag)
-  end
-
-  private
-
-  def find_tag
-    @tag = ActsAsTaggableOn::Tag.find_by(name: params[:name])
-  end
+  # before_action :find_tag, :only => [:index]
```
bboe: Should we delete this controller?

test/controllers/pictures_controller_test.rb
```diff
@@ -3,8 +3,8 @@
 class PicturesControllerTest < ActionController::TestCase
   setup do
     @pictures = [
-      Picture.create(link: 'http://www.foo.bar/img.jpg', tag_list: []),
-      Picture.create(link: 'http://www.foo.bar/img.jpg', tag_list: [])
+      Picture.create(link: 'http://www.foo.bar/img1.jpg', tag_list: ['tag1']),
```
bboe: What would happen if `create` happened to fail? How can we ensure that if it does fail we know about it?

test/controllers/pictures_controller_test.rb
```diff
@@ -62,4 +62,11 @@ class PicturesControllerTest < ActionController::TestCase
     assert_redirected_to root_path
     assert_equal 'Delete Success!', flash[:notice]
   end
+
+  test "can show images by tag" do
+    picture = @pictures.last
+    get :index, tag: picture.tag_list[0]
```
bboe: What's another way to get the list of tags other than through a Picture?

bboe: What's another way to get the list of tags other than through a Picture?

test/controllers/pictures_controller_test.rb
```diff
@@ -62,4 +62,11 @@ class PicturesControllerTest < ActionController::TestCase
...
+    assert_response :success
+    assert_select '#picture_list .image', 2
```
bboe: If we have a "Tag" object in the test, can we more precisely know how many images should be associated with it?

test/controllers/tags_controller_test.rb
```diff
@@ -1,17 +1,4 @@
 require 'test_helper'
 
 class TagsControllerTest < ActionController::TestCase
```
bboe: Is this test needed anymore?

app/controllers/pictures_controller.rb
shiboying: Yes, you are right. I will change it to `@pictures = @params_tag ? Picture.tagged_with(@params_tag) : Picture.all.order('updated_at DESC')
`
For the ordering, there is no specification for that but I think it is good I make it order the pictures in the same way.

test/controllers/pictures_controller_test.rb
shiboying: A `!` is necessary :)

test/controllers/pictures_controller_test.rb
shiboying: Will fix.

test/controllers/pictures_controller_test.rb
shiboying: Good point! Will fix.

test/controllers/tags_controller_test.rb
shiboying: For tags controller and controller test, I keep it for the next story. Still need a tags#index to show all tags. 
But it is a good catch that we need to remove all unnecessary codes to keep it clean. Thanks!

app/views/pictures/index.html.erb
```diff
@@ -4,4 +4,6 @@
 
 <%= render partial: 'list/picture_list', object: @pictures %>
```
rmacklin: A few comments:

1. According to Rails conventions, having the `app/views/list/` directory implies there is a `ListController` in our app. There is no such controller here, and this partial is used in `pictures#index` so it would make the application file structure more intuitive/conventional if the `picture_list` partial was instead in `app/views/pictures/`. However...
2. ...This partial was extracted when had duplicated the `pictures#index` functionality in the `tag#show` action (based on https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/9#discussion_r50940086). However, this PR removes that duplication, and thus the need for this partial has gone away. However...
3. ...I still think we could use a partial. But, I think the one that makes the most sense is a partial for a single picture:
```erb
<li>
  <%= link_to (image_tag picture.link, class: 'image'), picture %>
  <% picture.tag_list.each do |tag_name| %>
    <%= link_to tag_name, tag_path(tag_name) %><br>
  <% end %>
</li>
```
which would change this line into:
```erb
<ul id="picture_list">
  <% @pictures.each do |picture| %>
    <%= render partial: 'picture', object: picture %>
  <% end %>
</ul>
```
which can be reduced to the shorthand form:
```erb
<ul id="picture_list">
  <% @pictures.each do |picture| %>
    <%= render picture %>
  <% end %>
</ul>
```
which is pretty clean. But we can still improve this one step further:
```erb
<ul id="picture_list">
  <%= render @pictures %>
</ul>
```
which is really nice and compact. (FYI, `render @pictures` is just equivalent shorthand for:
`render partial: 'picture', collection: @pictures`.
For more details, see http://guides.rubyonrails.org/layouts_and_rendering.html#using-partials.)

app/views/pictures/index.html.erb
```diff
@@ -4,4 +4,6 @@
...
 
-<%= link_to('Upload New Image', new_picture_path) %>
+<% if @params_tag.blank? %>
+  <%= link_to('Upload New Image', new_picture_path) %>
+<% end %>
```
rmacklin: Interesting. I don't necessarily think we need to hide this button if we are viewing images filtered by a tag. In fact, perhaps even better would be:
- Always render the button
- If there is a tag filter, change the button to "Upload new #mytag image" and have the button autofill "mytag" in the tag_list on the form.
You don't have to do that, but it could be a fun exercise. In either case, I would probably not hide the Upload button even if there is a filter applied.

config/routes.rb
```diff
@@ -5,7 +5,6 @@
   # You can have the root of your site routed with "root"
   root 'pictures#index'
   resources :pictures, only: [:new, :create, :index, :show, :destroy, :edit, :update]
-  resources :tags, only: [:show], param: :name
```
rmacklin: So, in https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/9#discussion_r50940086 we weren't suggesting to completely remove the `show` action. Rather, we thought the `show` action should be a simple page displaying the tag and maybe the number of images with that tag, which could be a link to the `pictures#index` filtered by that tag. It could also contain edit and delete links for the tag, but that can of course come later.

app/views/pictures/index.html.erb
shiboying: Reasonable and detailed and nice. I will follow your suggestion to change my code.

app/views/pictures/index.html.erb
shiboying: "Upload new #mytag image" is an interesting feature, maybe later I will add that. So, if we think that is a filter of all images by tag, yes I think we should render this button. Although what I thought before was to render a new page to show those images not just filter it. But I think regarding it as a filter on index page is reasonable and acceptable, I will change it to always render that button.

config/routes.rb
shiboying: Yes, I know that we have discussed that we should have some functionalities for `show` action. The reason why I deleted it is just like you said, I wanted to make it later because it is not necessary to implement in this story.  And I plan to finish the next "show all tags" story first and then add the `show` action in tags controller. So I don't want to make it look confused if I keep `show` action but without implementing anything in the "show all tags" story.

rmacklin: Makes sense. Thanks for explaining! :+1:

app/controllers/pictures_controller.rb
```diff
@@ -1,8 +1,10 @@
 class PicturesController < ApplicationController
-  before_filter :find_picture, :only => [:update, :show, :edit, :destroy]
+  before_action :find_picture, :only => [:update, :show, :edit, :destroy]
 
   def index
-    @pictures = Picture.all.order('updated_at DESC')
+    @params_tag = params[:tag]
```
rmacklin: See https://github.com/AppFolioOnboarding/Boying_ImageUploader/commit/8cf44af2ee6c6bcccdcf9ed859df81367fe9db91#commitcomment-15746386

app/views/pictures/index.html.erb
rmacklin: :+1:

app/controllers/pictures_controller.rb
shiboying: Yes, you are right, I will fix it.

app/controllers/pictures_controller.rb
```diff
@@ -1,8 +1,10 @@
 class PicturesController < ApplicationController
-  before_filter :find_picture, :only => [:update, :show, :edit, :destroy]
+  before_action :find_picture, :only => [:update, :show, :edit, :destroy]
 
   def index
-    @pictures = Picture.all.order('updated_at DESC')
+    tag = params[:tag]
+    pictures = tag ? Picture.tagged_with(tag) : Picture.all
```
rmacklin: This is a bit of an edge case, but if someone were to go to the path: `/pictures?tag=` (i.e. pass an empty string or a blank string for the tag, we would try to look up that pictures with that tag. It may make more sense to ignore that if the tag param is blank. You could do that by changing the ternary expression from `tag` to `tag.present?`.

As I said it's an edge case so only change it if you feel like it.

shiboying: Yes, that's right, I will change it. And I think I can also add a something to prevent access to `/pictures?tag=nosuchtag`. And I plan to redirect to the root page and show a notice for that.

shiboying:
@rmacklin Do you have any commits on my new commit?

rmacklin:
@shiboying Yes, I left my comments directly on the commit: https://github.com/AppFolioOnboarding/Boying_ImageUploader/commit/7d3d28f48cdc97daa7225f900505eb603646630f#commitcomment-15751341

shiboying:
@rmacklin Since I amended the fix of invalid HTML and force pushed, my reply to your comments cannot be found. Good to know about this and I will pay attention to this kind of things in the future.

rmacklin:
@shiboying no worries. Your comment still exists here: https://github.com/AppFolioOnboarding/Boying_ImageUploader/commit/d13d0abd8a1b0d4560238ec66ba9bd19036dd70d#commitcomment-15815550.

I think amending the commit was a good idea. In the future, before you force push the branch you could copy the link to your comment that's on the individual commit and then paste that link in a comment on the PR (like I did above).

shiboying:
@rmacklin Got it!
Can I merge it to master?

rmacklin:
It looks good on my end, but I didn't check the deployed review on. If you verified the app is working properly there, then you can go ahead and merge.

shiboying:
I verified it is working well on the deployed review app. I am going to merge it.

PR #24, opened 2016-02-02
Tags index page

shiboying:


shiboying:
After checking on the deployed review app, I found it would show a blank page(only with title) if there are no tags saved. I was just about to deal with this edge case and then I remembered the following story is "As a user I want all images to have at least one tag". So, I think after finishing that story, there will be no such condition. Therefore, I decide not to deal with this case in this story.

shiboying:
But there is still a case that if no picture is uploaded.Then I still need to deal with it. I will add a conditional check for it.

rmacklin:
Great job thinking about edge cases!

app/views/tags/index.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1>Show All Tags</h1>
+
+<table id="tags_list">
+  <% @tags.each do |tag| %>
+    <tr>
+      <td>Tag: <%= link_to tag.name, pictures_path(tag: tag.name) %></td>
```
rmacklin: Based on the discussion in https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/10#discussion_r51185866, I think ultimately the tag index page should link to the show page for each tag, rather than to the pictures index filtered by that tag (which could instead be linked from the tag show page).

After rereading that discussion, I see that you had decided to complete the tag index page before the show page, which means you cannot link to the tag show page from the tag index page right now. I think that's okay, but are you planning to come back and change this after you add the tag show page?

Also, this made me realize that we do not have a story to add a tag show page anymore (because the story which used to be about implementing a tag show page was reworked as the pictures index with a tag query parameter). We should add this story and consider making it a prerequisite of the tag index story.

config/routes.rb
```diff
@@ -4,8 +4,8 @@
 
   # You can have the root of your site routed with "root"
   root 'pictures#index'
-  resources :pictures, only: [:new, :create, :index, :show, :destroy, :edit, :update]
-
+  resources :pictures
+  resources :tags, only:[:index]
```
rmacklin: This isn't a big deal because we will soon be adding another action to this array, but I believe when you only have one element you can omit the array (i.e. `only: :index` instead of `only: [:index]`) because Rails will wrap it in an array for you.

app/views/tags/index.html.erb
shiboying: Yes, I am just thinking about tags#show. And just not sure where I can link to this show page. I think you are right, this place is a good place to link to the show page. 

And here is the summary of what you have mentioned before about what should be shown on this `show` page, 

> we thought the show action should be a simple page displaying the tag and maybe the number of images with that tag

>This page could have edit and delete links for the tag resource. It could include a link to pictures#index filtered by that tag (even going as far as including a preview of a few of the images with that tag)

I think they can be treated as acceptance criteria for the story and adding a story for that could be more helpful.

Yes, I plan add this after this story, and I think a new branch for that is better? Or add it in this branch? What do you think?

config/routes.rb
shiboying: Yes, you are right but since a `show` is coming soon, I will not change it now. Good point, thanks!

app/views/tags/index.html.erb
```diff
@@ -0,0 +1,10 @@
...
+      <td>Count: <%= tag.taggings_count %></td>
```
rmacklin: Is there another way we could indicate that the previous cell is the _Tag_ and that this cell is the _Count_?

test/controllers/tags_controller_test.rb
```diff
@@ -1,4 +1,25 @@
 require 'test_helper'
 
 class TagsControllerTest < ActionController::TestCase
+  test 'can show all tags order by frequency' do
+    Picture.create!(link: 'http://www.foo.bar/img1.jpg', tag_list: ['tag1'])
+    Picture.create!(link: 'http://www.foo.bar/img2.jpg', tag_list: ['tag1', 'tag2'])
+    Picture.create!(link: 'http://www.foo.bar/img2.jpg', tag_list: ['tag1', 'tag2', 'tag3'])
+
+    @tags = ActsAsTaggableOn::Tag.all.order(taggings_count: :desc)
```
rmacklin: Is it necessary to use `ActsAsTaggableOn` in this test? Can you think of why we may want to avoid that?

test/controllers/tags_controller_test.rb
```diff
@@ -1,4 +1,25 @@
...
+    top_tag = @tags.first
+    get :index
+
+    assert_response :success
+    assert_select '#tags_list tr', ActsAsTaggableOn::Tag.count
+    assert_select '#tags_list tr:first-child td:first-child', "Tag: #{top_tag.name}"
+    assert_select '#tags_list tr:first-child td:last-child', "Count: #{top_tag.taggings_count}"
+  end
+
+  test 'redirect to root page if no pictures saved' do
```
rmacklin: Nice job testing the edge case!

test/controllers/tags_controller_test.rb
```diff
@@ -1,4 +1,25 @@
...
+    get :index
+
+    assert_redirected_to root_path
```
rmacklin: Perhaps we should add a flash notice?

app/views/tags/index.html.erb
rmacklin: Yeah, a new branch would definitely make sense.

As for the acceptance criteria, I wouldn't include "edit and delete links" as part of the show page acceptance criteria, because we won't have implemented edit and delete at that point.

test/controllers/tags_controller_test.rb
shiboying: So if there are no pictures shown, the root page will display a sentence "Please upload some images first.". So when clicking on "Show All Tags", it will come back to this page and shown "Please upload some images first.". So I think that one is enough for the notice of no pictures and no tags. Do you think it is enough?

rmacklin: OK, that's fine.

app/views/tags/index.html.erb
shiboying: Ok, I will do it later.

app/views/tags/index.html.erb
shiboying: You mean a different way than `assert_select '#tags_list tr:first-child td:first-child', "Tag: #{top_tag.name}"` in test?

test/controllers/tags_controller_test.rb
shiboying: I didn't find out the reason of avoiding using that...Could you give me more hints?

app/views/tags/index.html.erb
rmacklin: No, that's not what I mean. You are using a table to display this information. Is there a feature of tables that could let us omit printing "Tag:" and "Count:" in each row?

test/controllers/tags_controller_test.rb
rmacklin: Based on the pictures you have created above, we already know that we expected the tag index to look like (in the form `<tag-name> (<tag-count>)`):
> tag1 (3)
> tag2 (2)
> tag3 (1)

so we can make our assertions based on that alone. It's not necessary (and it introduces coupling) to use `ActsAsTaggableOn::Tag.all.order(taggings_count: :desc)` here. The test does not need to know that our implementation of `Tag`s uses `ActsAsTaggableOn`.

app/views/tags/index.html.erb
shiboying: Oh, I see. I think I can use:
```ruby
  <tr>
    <th>Tag</th>
    <th>Count</th>
  </tr>
```
Is that right?

test/controllers/tags_controller_test.rb
shiboying: I see. Is that the same case with 
`assert_select '#tags_content tr', ActsAsTaggableOn::Tag.count`? I should use `3` instead?

And if sometime in the future someone change the test case, we need to manually change the assertions. Does it matter?

shiboying:
Rebase on master (Since I merged `Delete confirmation` PR). 
And I added a commit `Add table header for tags list and change the way for testing the order of tags`

test/controllers/tags_controller_test.rb
shiboying: I changed a new way to test the order in the latest commit.

app/views/tags/index.html.erb
rmacklin: Yes, precisely.

app/views/tags/index.html.erb
```diff
@@ -0,0 +1,16 @@
+<h1>Show All Tags</h1>
+
+<table id="tags_list">
+  <tr>
+    <th>Tag</th>
+    <th>Count</th>
```
rmacklin: :+1: Nice job using the `th`s here.

app/views/tags/index.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  </tr>
+  <div id="tags_list_content">
```
rmacklin: This is invalid HTML (in addition to being not being semantic). Instead you could put the `<tr>` above inside a semantic `<thead>` and change this `<div>` into a semantic `<tbody>` element.

test/controllers/tags_controller_test.rb
```diff
@@ -1,4 +1,22 @@
...
+    Picture.create!(link: 'http://www.foo.bar/img1.jpg', tag_list: ['tag1'])
+    Picture.create!(link: 'http://www.foo.bar/img2.jpg', tag_list: ['tag1', 'tag2'])
+    Picture.create!(link: 'http://www.foo.bar/img2.jpg', tag_list: ['tag1', 'tag2', 'tag3'])
+
+    get :index
+
+    assert_response :success
+    assert_select '#tags_list_content tr td' do |elements|
+      assert_equal %w(tag1 3 tag2 2 tag3 1), elements.map { |el| el.children.text }
+    end
```
rmacklin: This is a definite improvement by not having this test be tied to `ActsAsTaggableOn`. Nice job :+1:

That said, this line:
```ruby
assert_equal %w(tag1 3 tag2 2 tag3 1), elements.map { |el| el.children.text }
```
feels a bit dirty. I wonder if we could refactor the assert_select to separate the assertions on the "count" `td` from the "tag" `td`.

app/views/tags/index.html.erb
shiboying: @bboe  shall we make a story for tags#show?

PR #25, opened 2016-02-02
bs - Add delete confirmation.

shiboying:
There is a new acceptance criteria - 'The user is prompted for confirmation prior to deletion.' in my issues.


app/views/pictures/show.html.erb
```diff
@@ -13,4 +13,4 @@
 
 <%= link_to 'Go Back', root_path %>
 <%= link_to 'Edit it', edit_picture_path %>
-<%= link_to 'Delete it', picture_path, method: :delete %>
+<%= link_to 'Delete it', picture_path, method: :delete, data: { confirm: 'Are you sure?' } %>
```
rmacklin: Looks good. This behavior is something we'll want to test with capybara when we add a test for the "delete a picture" flow.

shiboying: Yes, will test it later.

PR #27, opened 2016-02-03
Edit picture

### LINKED ISSUE #21: Modify Tag List
#### As a user I want to be able to edit the tag list for an image.

Acceptance criteria:
- [ ] The image whose tags are to be modified is displayed on the edit page.

shiboying:
This PR refers to: closes #21
Since I previously added edit action and page, this branch is just to finish edit feature.


bboe:
In the pull request description above can you add a reference to the issue that this pull request refers to? A great way to do that is via, "Closes #<NUMBER>.", where `<NUMBER>` is the issue number. With this in place it's easy to see what the purpose of this pull request is, and when looking at the list of issues to quickly find the pull request that is associated with the issue. Also, github will automatically close that issue when this pull request is merged if it's written like I suggested.

test/controllers/pictures_controller_test.rb
```diff
@@ -34,9 +34,33 @@ class PicturesControllerTest < ActionController::TestCase
 
     assert_redirected_to root_path
     picture.reload
+    assert_equal 'Update Success!', flash[:notice]
```
bboe: I know that you created the `edit` action in another PR. However, I don't see any tests for the edit action. What tests are needed for it?

test/controllers/pictures_controller_test.rb
```diff
@@ -54,6 +78,14 @@ class PicturesControllerTest < ActionController::TestCase
     assert_select '#image_errors li', 1
   end
 
+  test 'multiple tags displayed correctly with comma on error' do
```
bboe: Should we have a similar test for on update?

test/controllers/pictures_controller_test.rb
shiboying: Functional test to test we can get edit successfully.

test/controllers/pictures_controller_test.rb
shiboying: Yes, I think that is more complete.

shiboying:
@bboe I have modifed my codes according to your comments. Thanks!

lzhang0616:
I have QA your review app! It looks great!

test/controllers/pictures_controller_test.rb
```diff
@@ -27,6 +27,13 @@ class PicturesControllerTest < ActionController::TestCase
     end
   end
 
+  test 'can access edit' do
```
bboe: This is great. What about when trying to edit an image that doesn't exist?

app/views/pictures/_validation_errors.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  <div id="image_errors">
+    <h2><%= pluralize(@picture.errors.count, 'error') %> prohibited this image from being saved:</h2>
+
+    <ul>
+      <% @picture.errors.full_messages.each do |message| %>
+        <li><%= message %></li>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
```
rmacklin: This is cool, though it made me realize @aaronpanch did not introduce you to `simple_form`. I guess he only showed it to Haomin. We use the `simple_form` gem in our rails apps because it provides some convenient features on top of the regular `form_for` method, including inline validation errors. You don't have to change to using `simple_form` here if you don't want to, but if you would like to try it, maybe Haomin can show you how he set it up.

test/controllers/pictures_controller_test.rb
```diff
@@ -27,6 +27,13 @@ class PicturesControllerTest < ActionController::TestCase
...
+    picture = create_pictures.first
+    get :edit, id: picture.id
+
+    assert_response :success
```
rmacklin: Perhaps we can add some kind of `assert_select` assertion about the content of the edit page?

test/controllers/pictures_controller_test.rb
```diff
@@ -34,9 +41,36 @@ class PicturesControllerTest < ActionController::TestCase
...
     assert_equal %w(tag1 tag2 tag3), picture.tag_list
   end
 
+  test 'can update picture url' do
+    picture = create_pictures.first
+    current_link = picture.link
+    new_link = 'http://thisis.anewlink.com'
+    patch :update, id: picture.id, picture: { link: new_link }
+
+    assert_not_equal new_link, current_link
```
rmacklin: I'm not sure I see the value of this assertion.

test/controllers/pictures_controller_test.rb
```diff
@@ -34,9 +41,36 @@ class PicturesControllerTest < ActionController::TestCase
...
+    assert_redirected_to root_path
+    picture.reload
```
rmacklin: Not a big deal, but you could inline this into line 58 (`picture.reload.link`).

test/controllers/pictures_controller_test.rb
```diff
@@ -34,9 +41,36 @@ class PicturesControllerTest < ActionController::TestCase
...
+    assert_equal new_link, picture.link
+  end
+
+  test 'cannot update picture if input invalid' do
+    picture = create_pictures.first
+    current_link = picture.link
+    new_link = 'abc'
+    patch :update, id: picture.id, picture: { link: new_link, tag_list: '' }
+
+    assert_not_equal current_link, new_link
```
rmacklin: Same as above: I'm not sure I see the value of this assertion.

rmacklin:
@lzhang0616 Did you confirm the acceptance criteria?

test/controllers/pictures_controller_test.rb
shiboying: Yes, that is a bug in my app. I'll fix it.

app/views/pictures/_validation_errors.html.erb
shiboying: That needs a lot changes in this but I will use it under `WebsiteStyle` branch.

test/controllers/pictures_controller_test.rb
shiboying: Yes, of course.

test/controllers/pictures_controller_test.rb
shiboying: For `can update` case, actually we can allow updating a same value, so I think we do not need to `assert_not_equal` for that, and assert redirection is enough. And for `cannot update`, we only care about the update result remain the same. And assert error messages and the flash notice is enough for that. How do you think? @bboe

shiboying:
@rmacklin @lzhang0616 Yes, I will add to show this picture on the edit page.

test/controllers/pictures_controller_test.rb
shiboying: Sure, I will change it.

test/controllers/pictures_controller_test.rb
bboe: My thought process is that the test for updating successfully depends on the value being changed. The test itself does not make it clear what the original value was, and so there is no guarantee that the value that we want to set is different than what is already set. It's possible, albeit unlikely, that someone modifies create_pictures such that it's initial value is the same as the value we want to update it to. However, many things we test for are unlikely, and we want to ensure we don't introduce a regression. Regressions can occur both in the code, and in the test code.

In the failing update case, we needn't actually ensure the value we want to update to is different than what we try to set. We can instead ensure that the value we tried to update to is not the value after the failed update. In that case we do not care what the original value is.

rmacklin: What do you think about:
```ruby
current_link = picture.link
new_link = "#{current_link}abc"
```
That guarantees that the two are different without an assertion. The reason I prefer it is that if I am trying to understand some code by looking at the test, I will look at all the assertions. So I like limiting assertions to be testing the output of the system under test (by contrast this assertion is testing the input).

bboe: :+1: That's a great way to ensure the value is changed.

shiboying: Cool, I will make this change in a separate commit in capybara test branch.

bboe: @shiboying awesome. However, might I suggest making a separate pull request with just that change? One can never have too many pull requests.

shiboying: Oh, yes. I will do that.

PR #41, opened 2016-02-05
ensure current_link is not equal to new_link

shiboying:


shiboying:
@bboe This is the pull request for changing the assertion of `current_link` and `new_link`.

test/controllers/pictures_controller_test.rb
```diff
@@ -120,10 +120,10 @@ class PicturesControllerTest < ActionController::TestCase
   test 'can update picture url' do
     picture = create_pictures.first
     current_link = picture.link
-    new_link = 'http://thisis.anewlink.com'
+    new_link = "http://thisis.anewlink.com/#{current_link}"
```
bboe: Will this have two `http://` portions inside of it?

test/controllers/pictures_controller_test.rb
```diff
@@ -140,10 +140,10 @@ class PicturesControllerTest < ActionController::TestCase
   test 'cannot update picture if input invalid' do
     picture = create_pictures.first
     current_link = picture.link
-    new_link = 'abc'
+    new_link = "abc/#{current_link}"
```
bboe: In this case I don't think `new_link` needs to derive from current_link (that will work though). Rather, we can assert that the update didn't happen (replace 148 with `assert_not_equal`).

test/controllers/pictures_controller_test.rb
shiboying: Yes, I see. A ```new_link = "#{current_link}/abc" seems better to form a real new link.

test/controllers/pictures_controller_test.rb
shiboying: like replay it with`assert_not_equal new_link, picture.reload.link`?

bboe: Yes.

shiboying:
@bboe I have a new commit for that.

PR #46, opened 2016-02-11
Content-Security-Policy

shiboying:


app/controllers/application_controller.rb
```diff
@@ -2,4 +2,9 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
+  before_action :set_csp
```
bboe: This is one great way to approach setting the headers. There is another way which I would argue is preferred over doing it as part of the before chain. Check out `default_headers` at http://edgeguides.rubyonrails.org/configuring.html#configuring-action-dispatch for an example.

bboe:
This story is more interesting once you have merged in your CSS changes. I would suggest finishing that up first, and then rebasing this branch and seeing if things still work the way you expect.

app/controllers/application_controller.rb
shiboying: I'll take a look, thanks.

shiboying:
@bboe Yes, I agree. Since there is not much content style shown. I will first use the other way to add this configuration and then wait until that branch merged.

shiboying:
@bboe I rebased on my website style branch and fixed inline style, there are no console errors. I think it can be merged. How do you think?

bboe:
Looks good. My only question is what purpose does tether have?

shiboying:
I modified the commit message for adding tether. If I don't require tether, it will have a console error.

PR #51, opened 2016-02-18
Share image

### LINKED ISSUE #23: Share Image
#### As a user I want to share an image with someone via email.

Acceptance criteria:
- [x] There is a link to share an image.
- [x] After clicking on this link, a bootstrap modal appears with a form to
  send the email.
- [x] The modal closes and a success message is shown if the email was sent
  successfully.
- [x] The user is alerted of errors if the operation was unsuccessful.

Discussion topics:
- [ ] How could we develop the front end and back end separately? (And why
  might that be a good idea?)
- [ ] How should we implement a non-CRUD action?
- [ ] What are the different types of errors we should handle?
- [ ] What is AJAX? What libraries can help us add AJAX to our Rails app?
- [ ] How should we go about sending emails from our Rails app?
- [ ] How could we show a preview of the image being shared in the modal?
- [ ] What are the implications of allowing users to send emails from our app?

Reference material:
- http://v4-alpha.getbootstrap.com/components/modal
- http://guides.rubyonrails.org/action_mailer_basics.html
- https://elements.heroku.com/addons/sendgrid
- https://devcenter.heroku.com/articles/sendgrid#ruby-rails
- https://github.com/rails/jquery-ujs/wiki

Dependencies:
- ES6 configuration

shiboying:
closes #23


shiboying:
@rmacklin Please help me review this : )

app/assets/javascripts/images/index/sharing_modal.js
```diff
@@ -0,0 +1,41 @@
...
+  $(document).on('ajax:success', '#emailModal', (e, xhr, status, error) => {
+    $('#emailModal').modal('hide');
+    $('.site-content > .container').prepend(xhr.flash_message_html);
+  });
+
+  $(document).on('ajax:error', '#emailModal', (e, xhr, status, error) => {
+    switch (xhr.status) {
+      case UNPROCESSABLE_ENTITY:
+        $('#emailModalBody').html(xhr.responseJSON.form_html);
+        alert('Please enter valid input!');
```
rmacklin: This alert may not be necessary given that the user will see the validation errors in their modal as a result of the previous line.

app/assets/javascripts/images/index/sharing_modal.js
```diff
@@ -0,0 +1,41 @@
...
+        break;
+      case NOT_FOUND:
+        alert('Image not found!');
```
rmacklin: Could we provide more detail to the user in this error message? What scenario could cause a not found response?

app/assets/javascripts/images/index/sharing_modal.js
```diff
@@ -0,0 +1,41 @@
...
+        break;
+      default:
+        alert('Sharing failed due to some reason: ' + xhr.status);
```
rmacklin: Technical details such as the xhr status do not provide value to end users, and in some cases can be considered harmful because the error message could seem intimidating (the user might think: "am I supposed to know what this means and how to fix it?"). So I'd recommend removing the status.

Additionally, consider using a more user-friendly tone (e.g. "We're sorry, something went wrong. Please try again.").

app/views/pictures/_share_form.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= simple_form_for share_form,
+                    as: :share_form,
+                    remote: true,
+                    url: '', # url is required by simple form but we use data-path-template to set form action
+                    html: { 'data-path-template': share_picture_path('ID_PLACEHOLDER') } do |f| %>
+  <%= f.input :picture_link, as: :hidden %>
```
rmacklin: I'm not sure this hidden input is necessary. The server should be able to retrieve the picture link from the `Picture` model associated with this `share_picture` request.

app/assets/javascripts/images/index/sharing_modal.js
```diff
@@ -0,0 +1,41 @@
...
+
+  $(document).on('show.bs.modal', '#emailModal', (event) => {
+    const $shareButton = $(event.relatedTarget);
+    const $shareForm = $('#new_share_form');
+    const pictureId = $shareButton.closest('[data-picture-id]').data('picture-id');
+    const formAction = $shareForm.data('path-template').replace('ID_PLACEHOLDER', pictureId);
+    const sharedPictureLink = $shareButton.data('url');
+
+    $shareForm.attr('action', formAction);
+    $('#share_form_picture_link').val(sharedPictureLink);
```
rmacklin: Based on https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/51#discussion_r53519438, I think you'll be able to remove this line.

app/assets/javascripts/images/index/sharing_modal.js
```diff
@@ -0,0 +1,41 @@
...
+    $('#picture-preview').attr('src', sharedPictureLink);
+  });
+
+  $(document).on('hidden.bs.modal', '#emailModal', () =>$('#emailModal').replaceWith($blankModal.clone()));
```
rmacklin: Minor, but could you put a space after the `=>` for readability?

rmacklin: You don't have to change the implementation if you don't want to, but we discovered a simpler way to do this. Instead of using `replaceWith($blankModal.clone())` you could use `.html($blankModal.html())` (which will only replace the inner HTML of the `#emailModal` element, not the node itself.

The benefit of doing this is that the `#emailModal` element will never be destroyed, which means you can save a reference to `$('#emailModal')` are attach all of your event listeners to that, instead of attaching them to `$(document)` and filtering by `'#emailModal'`. This is a bit more efficient as you only have to do one jQuery lookup (and then reuse that jQuery element everywhere) instead of multiple lookups with event filtering.

app/views/pictures/_sharing_modal.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="modal fade" id="emailModal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
```
rmacklin: What do you think about the name `shareModal` instead of `emailModal`? It's a bit more specific. Additionally, it better matches the name of this partial (`sharing_modal`...which could also be renamed to `share_modal` if you want them to have an exact match).

app/assets/javascripts/images/index/sharing_modal.js
```diff
@@ -0,0 +1,41 @@
+const NOT_FOUND = 404;
+const UNPROCESSABLE_ENTITY = 422;
+
+function sharing() {
+  const $blankModal = $('#emailModal').clone();
```
rmacklin: This module does not need to be concerned with the selector of the modal element (`'#emailModal'`). It would be better not to hardcode that here, because then this module could be more easily reused on another page with a different selector. Can you think of a way to extract that knowledge out of this module?

app/assets/javascripts/images/index/main.js
```diff
@@ -1,3 +1,3 @@
-import hello from './hello'
+import sharing from './sharing_modal'
 
-hello()
+sharing();
```
rmacklin: The name of this method is a bit vague. What do you think about renaming it to something that's more descriptive of what the method actually does?

app/assets/stylesheets/_footer.scss
```diff
@@ -1,5 +1,14 @@
...
   background: #f7f7f9;
+  bottom: 0;
+  height: 180px;
+  position: absolute;
+  width: 100%;
+
+  & > .container {
+    padding-left: 15px;
+    padding-right: 15px;
+  }
```
rmacklin: I'm going to let @aaronpanch comment on the other CSS changes in this PR. But I'm curious: why did you change the footer styles in this PR? Was it related to implementing the share image functionality?

app/controllers/pictures_controller.rb
```diff
@@ -1,7 +1,18 @@
 class PicturesController < ApplicationController
-  before_action :find_picture, :only => [:update, :show, :edit, :destroy]
+  before_action :find_picture, :only => [:update, :show, :edit, :destroy, :share]
+  rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
+
+  class ShareForm
+    include ActiveModel::Model
+    attr_accessor :email, :message, :picture_link
+
+    validates :picture_link, presence: true
```
rmacklin: Based on https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/51#discussion_r53519438, I think you'll be able to remove this line.

app/controllers/pictures_controller.rb
```diff
@@ -41,6 +52,21 @@ def destroy
...
+      PictureMailer.sharing_email(@share_form).deliver_now
+      flash.now[:success] = 'Share Successfully!'
+
+      flash_message_html = render_to_string(partial: 'layouts/flash_message')
+      render json: {flash_message_html: flash_message_html}
+    else
+      form_html = render_to_string(partial: 'share_form', object: @share_form)
+      render json: { form_html: form_html }, status: :unprocessable_entity
+    end
+  end
```
rmacklin: :+1: This is great! It's very easy to read and understand this action.

app/controllers/pictures_controller.rb
```diff
@@ -1,7 +1,18 @@
...
-  before_action :find_picture, :only => [:update, :show, :edit, :destroy]
+  before_action :find_picture, :only => [:update, :show, :edit, :destroy, :share]
+  rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
+
+  class ShareForm
+    include ActiveModel::Model
+    attr_accessor :email, :message, :picture_link
+
+    validates :picture_link, presence: true
+    validates :email, presence: true, format: { with: /\A[^@\s]+@([^@\s]+\.)+[^@\s]+\z/ }
```
rmacklin: Minor, but (for readability) consider extracting this regex to a constant named after what it represents.

app/controllers/pictures_controller.rb
```diff
@@ -48,7 +74,14 @@ def picture_params
   end
 
   def find_picture
-    @picture = Picture.find_by_id(params[:id])
-    redirect_to root_path, flash: { danger: 'No such image!' } if @picture.blank?
+    @picture = Picture.find(params[:id])
   end
+
+  def record_not_found(e)
```
rmacklin: The parameter `e` appears to be unused.

Just curious, are you running pronto on your changes? If so, did it complain about the unused parameter here?

app/mailers/application_mailer.rb
```diff
@@ -0,0 +1,4 @@
+class ApplicationMailer < ActionMailer::Base
+  default from: 'boying-image@appfolio.com'
```
rmacklin: This isn't a real email, is it? You could use `noreply@appfolio.com`.

app/controllers/pictures_controller.rb
```diff
@@ -1,7 +1,18 @@
 class PicturesController < ApplicationController
-  before_action :find_picture, :only => [:update, :show, :edit, :destroy]
+  before_action :find_picture, :only => [:update, :show, :edit, :destroy, :share]
+  rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
+
+  class ShareForm
+    include ActiveModel::Model
+    attr_accessor :email, :message, :picture_link
```
rmacklin: Consider renaming `email` to `recipient_email`.

app/views/layouts/_flash_message.html.erb
```diff
@@ -0,0 +1,8 @@
+<% flash.each do |message_type, message| %>
+  <div class="alert alert-<%= message_type %>">
+    <%= message %>
+    <button type="button" class="close" data-dismiss="alert" aria-label="Close"> 
+      <span aria-hidden="true">&times;</span> 
+    </button>
+  </div>
+<% end %>
```
rmacklin: :+1: Great job extracting this for DRY reuse.

app/views/picture_mailer/sharing_email.html.erb
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
```
rmacklin: It looks like this will be rendered within the `mailer` layout (based on [this line](https://github.com/AppFolioOnboarding/Boying_ImageUploader/blob/da851412eafa8d05efb589f9f9fb14175a496d2d/app/mailers/application_mailer.rb#L3)), which already includes the `<html>` and `<body>` tags. So I think you'll end up with those elements duplicated.

app/views/picture_mailer/sharing_email.html.erb
```diff
@@ -0,0 +1,13 @@
...
+    <p>
+      Your email is: <%= @recipient %>.<br>
```
rmacklin: I'm unsure of the value of this line. Can you share with me why you added it?

app/views/picture_mailer/sharing_email.html.erb
```diff
@@ -0,0 +1,13 @@
...
+      The message is: <%= @message %>.<br>
+      Picture shared with you: <%= image_tag @picture_link %>.
```
rmacklin: Remember, this email is going to go out to real people. Perhaps there is some friendlier wording we could use in this email body?

app/views/pictures/_picture.html.erb
```diff
@@ -1,8 +1,13 @@
...
-  <div class="card">
+  <div class="card" data-picture-id="<%= picture.id %>">
     <%= link_to (image_tag picture.link, class: 'image'), picture %>
-    <div class="js-tag-list card-block">
-      <%= render partial: 'tag', collection: picture.tag_list %>
+    <div class="js-tag-list card-block align-horizontal">
+      <div class="tag-list">
+        <%= render partial: 'tag', collection: picture.tag_list %>
+      </div>
+      <div>
```
rmacklin: A div with no attributes is typically a code smell.

app/views/pictures/_picture.html.erb
```diff
@@ -1,8 +1,13 @@
...
+        <%= link_to 'Share', '#', class: 'btn btn-primary btn-sm', 'data-toggle': 'modal', 'data-target': '#emailModal', 'data-url': "#{picture.link}" %>
```
rmacklin: Perhaps `data-picture-src` is a better name for this than `data-url`

app/views/pictures/_share_form.html.erb
```diff
@@ -0,0 +1,14 @@
...
+  <%= image_tag @picture.present? ? @picture.link : '', id: 'picture-preview', class: 'image' %>
+  <%= f.input :email %>
+  <%= f.input :message, as: :text %>
+  <div class="modal-footer">
+    <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
+    <%= f.submit 'Send', class: 'btn btn-primary', id: 'send-button'%>
```
rmacklin: What is this `id` used for?

app/views/pictures/_sharing_modal.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+        <h4 class="modal-title" id="myModalLabel">Email this picture</h4>
```
rmacklin: I think we should rename `myModalLabel`

app/views/tags/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="container m-t-1">
+<div class="m-t-1">
```
rmacklin: Most of these view changes seem unrelated to this Share Image PR

config/application.rb
```diff
@@ -25,7 +25,8 @@ class Application < Rails::Application
     ActsAsTaggableOn.remove_unused_tags = true
     config.browserify_rails.commandline_options = '-t babelify'
     config.action_dispatch.default_headers = {
-           'Content-Security-Policy' => "default-src 'none'; script-src 'self'; img-src *; style-src 'self'; connect-src 'self';"
+           'Content-Security-Policy' => "default-src 'none'; script-src 'self' 'unsafe-eval'; img-src *; style-src 'self'; connect-src 'self';"
```
rmacklin: We should look into this. If it's a bootstrap bug, we should file an issue on the bootstrap repo. (Even better, propose a PR if we can fix the bug.)

config/environment.rb
```diff
@@ -3,3 +3,13 @@
...
+
+ActionMailer::Base.smtp_settings = {
+  :address              => 'smtp.sendgrid.net',
+  :port                 => '587',
+  :authentication       => :plain,
+  :user_name            => ENV['SENDGRID_USERNAME'],
+  :password             => ENV['SENDGRID_PASSWORD'],
+  :domain               => 'heroku.com',
+  :enable_starttls_auto => true
+}
```
rmacklin: This seems like `production`-only config.

config/application.rb
```diff
@@ -25,7 +25,8 @@ class Application < Rails::Application
...
         }
+    ActionMailer::Base.delivery_method = :smtp
```
rmacklin: Is this line necessary?

app/views/pictures/_picture.html.erb
```diff
@@ -1,8 +1,13 @@
 <div class="js-picture col-xs-4">
```
rmacklin: Just curious: is this partial used anywhere other than the index page?

test/controllers/pictures_controller_test.rb
```diff
@@ -168,6 +168,35 @@ class PicturesControllerTest < ActionController::TestCase
     assert_equal 'Delete Successfully!', flash[:success]
   end
 
+  test 'can share an image with email' do
+    picture = create_pictures.first
+
+    assert_difference('ActionMailer::Base.deliveries.size', 1) do
+      post :share, id: picture.id, share_form: { email: 'abc@gmail.com', message: 'hi', picture_link: picture.link }, format: 'js'
```
rmacklin: In real life, we expect this request to be an `XHR`. Therefore, our tests should test it as an `xhr`.

rmacklin: Is `format: :js` necessary here? Additionally, you refactored the controller to send back `json` instead of `js` anyway, so it seems to be incorrect.

test/controllers/pictures_controller_test.rb
```diff
@@ -168,6 +168,35 @@ class PicturesControllerTest < ActionController::TestCase
...
+    end
+
+    assert_response :success
+    assert_equal 'Share Successfully!', flash[:success]
```
rmacklin: It would be better to directly assert that the response JSON includes this message.

test/controllers/pictures_controller_test.rb
```diff
@@ -168,6 +168,35 @@ class PicturesControllerTest < ActionController::TestCase
...
+  end
+
+  test 'cannot share an image with invalid input' do
+    picture = create_pictures.first
+
+    assert_no_difference('ActionMailer::Base.deliveries.size') do
+      post :share, id: picture.id, share_form: { email: 'invalid', message: 'hi', picture_link: picture.link }, format: 'js'
```
rmacklin: Same as above re: XHR and `format: 'js'`

test/controllers/pictures_controller_test.rb
```diff
@@ -168,6 +168,35 @@ class PicturesControllerTest < ActionController::TestCase
...
+    end
+
+    assert_response :unprocessable_entity
```
rmacklin: Can we assert anything about the JSON response body?

test/flow_tests/pictures_crud_test.rb
```diff
@@ -149,6 +149,48 @@ class PicturesCrudTest < FlowTestCase
...
+  test 'share a picture' do
+    create_picture
+
+    visit root_path
+    assert_not page.has_selector?('#new_share_form')
+
+    click_on('Share')
+    assert page.has_selector?('#new_share_form')
+
+    control_labels = page.all('.control-label')
```
rmacklin: It would be better to scope this lookup inside of the new_share_form.

test/flow_tests/pictures_crud_test.rb
```diff
@@ -149,6 +149,48 @@ class PicturesCrudTest < FlowTestCase
...
+    create_picture
+
+    visit root_path
+    assert_not page.has_selector?('#new_share_form')
+
+    click_on('Share')
+    assert page.has_selector?('#new_share_form')
+    control_labels = page.all('.control-label')
+    assert_equal '* Email', control_labels.first.text
+    assert_equal 'Message', control_labels.last.text
```
rmacklin: This is all repeated from the previous test. I'm not sure it needs to be in its own test anyway. We could first share the image with an invalid email, assert that we see validation errors, and then fix the email and successfully share the image all in one test.

test/flow_tests/pictures_crud_test.rb
```diff
@@ -149,6 +149,48 @@ class PicturesCrudTest < FlowTestCase
...
+    fill_in('* Email', with: 'iaminvalid')
+    fill_in('Message', with: 'This email has a message.')
+
+    click_on('Send')
+    assert_equal 'Please enter valid input!', page.driver.browser.switch_to.alert.text
+
+    page.driver.browser.switch_to.alert.accept
+    assert page.has_selector?('#new_share_form')
+  end
+
```
rmacklin: We are missing a test for the sharing a deleted image.

app/views/picture_mailer/sharing_email.html.erb
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <p>
+      Your email is: <%= @recipient %>.<br>
+      The message is: <%= @message %>.<br>
```
rmacklin: What will the email look like if the user did not provide a message? (This is a valid case because that field is optional.)

app/views/picture_mailer/sharing_email.text.erb
```diff
@@ -0,0 +1,3 @@
+Recipient <%= @recipient %>.
+Message: <%= @message %>.
+Picture: <%= @picture_link %>.
```
rmacklin: Same comments as above

test/mailers/picture_mailer_test.rb
```diff
@@ -0,0 +1,41 @@
...
+    share_form = create_share_form(picture, '')
+    email = PictureMailer.sharing_email(share_form).deliver_now
+
+    assert_not ActionMailer::Base.deliveries.empty?
+
+    assert_equal ['boying-image@appfolio.com'], email.from
+    assert_equal ['test@abc.com'], email.to
+    assert_equal 'An image has been shared with you', email.subject
+    assert_includes email.html_part.body.to_s, 'http://www.foo.bar/img1.jpg'
+    assert_includes email.text_part.body.to_s, 'http://www.foo.bar/img1.jpg'
```
rmacklin: These assertions are the same as the "with message" test. Which makes me wonder...why are there two tests?

rmacklin:
There are 31 commits in this PR. I did not look at them individually. I suspect that they could be squashed into a handful or two of atomic commits.

Additionally, some changes seem unrelated to the share image story (as I pointed out in a few of my inline comments). Those changes could be moved to another branch - an _atomic_ branch, analogous to an _atomic_ commit. This would make reviewing easier.

app/assets/javascripts/images/index/sharing_modal.js
shiboying: Yes, but it is weird my share_form cannot show validation errors below.. I will check with that.

app/assets/javascripts/images/index/sharing_modal.js
shiboying: When a user delete an image but the same time this image is being shared.

app/assets/javascripts/images/index/sharing_modal.js
shiboying: Yes, this information is not useful for end users. I didn't pay attention to this part and the alert text.

app/views/pictures/_share_form.html.erb
shiboying: Ok, i will try.

app/assets/javascripts/images/index/sharing_modal.js
shiboying: :+1:

app/assets/javascripts/images/index/sharing_modal.js
shiboying: Yes, I will try after cleaning my code

app/views/pictures/_sharing_modal.html.erb
shiboying: :+1:

app/assets/javascripts/images/index/sharing_modal.js
shiboying: @rmacklin if I extract `const $shareModal = $('#emailModal')` to line 3, it will not work since the replaceWith always change this part and I need `$shareModal` to point to latest current node but if I set it to `const` the pointer will always point to the old node. 

So I don't know if passing parameters work here and I don't know if I can pass parameters in this function.

app/assets/javascripts/images/index/main.js
shiboying: How about `share_picture`?

app/assets/stylesheets/_footer.scss
shiboying: Because previously I used another way to make footer fixed at bottom. Daniel helped me refactor it when pairing. I will move it to another branch.

app/controllers/pictures_controller.rb
shiboying: Sure.

app/controllers/pictures_controller.rb
shiboying: :v:

app/controllers/pictures_controller.rb
shiboying: Sure!

app/controllers/pictures_controller.rb
shiboying: No, I haven't use pronto before..

app/mailers/application_mailer.rb
shiboying: Ok, I will use that email.

app/controllers/pictures_controller.rb
shiboying: Sure.

app/views/layouts/_flash_message.html.erb
shiboying: :ghost:

app/views/picture_mailer/sharing_email.html.erb
shiboying: Good catch!

app/views/picture_mailer/sharing_email.html.erb
shiboying: ( ´◔ ‸◔') no specific reason.. actually just want to test I can pass correct values into the email. I will make a better sharing email template.

app/views/picture_mailer/sharing_email.html.erb
shiboying: Yes, I will change it.

app/views/pictures/_picture.html.erb
shiboying: They all have `class` attribute?

app/views/pictures/_picture.html.erb
shiboying: Sure.

app/views/pictures/_share_form.html.erb
shiboying: Oh, previously I used form and need to select this button in javascript. I will remove that.

app/views/pictures/_sharing_modal.html.erb
shiboying: Sure

app/views/tags/index.html.erb
shiboying: Yes, they are something related to the style.

config/application.rb
shiboying: Good.

config/environment.rb
shiboying: Ok, I will move it.

config/application.rb
shiboying: Yes, or I cannot send email and pass tests.

app/views/pictures/_picture.html.erb
shiboying: Yes, just used by index.

test/controllers/pictures_controller_test.rb
shiboying: Yes, it need to be an xhr.

test/controllers/pictures_controller_test.rb
shiboying: I will try.

test/controllers/pictures_controller_test.rb
shiboying: Good.

test/controllers/pictures_controller_test.rb
shiboying: Need to have a try.

test/flow_tests/pictures_crud_test.rb
shiboying: Sure.

test/flow_tests/pictures_crud_test.rb
shiboying: Sure.

test/flow_tests/pictures_crud_test.rb
shiboying: Yes, but I was thinking the ∆ for (unit test, controller test, integration test). I have tested I cannot share a deleted image (as well as i cannot share an image with invalid input) in controller test, so I just want to test the flow of sharing successfully and sharing unsuccessfully. 

The same idea is to test `create a picture`, I tested can create and cannot create in functional test. I didn't care about testing if it is picture_link invalid or tag_list is blank since I have tested them in unit test.

app/views/picture_mailer/sharing_email.html.erb
shiboying: `This message is:.` :eyes: 
I will change the email body.

app/views/picture_mailer/sharing_email.text.erb
shiboying: Yes.

test/mailers/picture_mailer_test.rb
shiboying: Just want to test if I can share it with or without message. I will change it the email body first.

shiboying:
@rmacklin Thanks for your reviews on my branch. 
They are detailed (`→_→……←_←……←_→……→_←……↓_↓……↑_↑……↓_↑……↖_↗……↙_↘`) and helpful, I took some notes based on them. And I will fix them based on your comments.

app/assets/javascripts/images/index/sharing_modal.js
rmacklin: Right. So we could say something like "Sorry, this image has been deleted."

app/assets/javascripts/images/index/sharing_modal.js
rmacklin: You can pass parameters to any function. You could pass the string `'#emailModal'` as a parameter to this function:

`sharing.js:`
```js
function sharing(modalSelector)
```
`main.js:`
```js
sharing('#emailModal')
```

app/assets/javascripts/images/index/main.js
rmacklin: Well, `share_picture` describes what happens _after you click share_. That's not exactly what the `sharing` method does. The `sharing` method attaches some event listeners for the sharing functionality.

app/assets/stylesheets/_footer.scss
rmacklin: OK, thanks. I will tell Daniel to make changes like that on a new branch too.

app/views/pictures/_picture.html.erb
rmacklin: Not the one here on line 8

app/views/pictures/_picture.html.erb
rmacklin: OK, thanks.

test/flow_tests/pictures_crud_test.rb
rmacklin: There is an important distinction.

Testing if `link` is invalid or if `tag_list` is invalid are both testing _validation errors_. The behavior in your code is not different when you `create` an image with a different kind of validation errors.

However, in the share image modal, there is different behavior when you try to share an image which is not found versus when you try to share an image with a validation error in the share form. So right now, we have no test covering [this part](https://github.com/AppFolioOnboarding/Boying_ImageUploader/blob/03b7657d4eb13d1b68c64fe14883b1688785602b/app/assets/javascripts/images/index/sharing_modal.js#L18-L20) of your javascript.

test/mailers/picture_mailer_test.rb
rmacklin: :+1:

app/assets/stylesheets/_footer.scss
shiboying: @rmacklin I found out the reason why we did refactor on container is try to use ` $('.site-content > .container').prepend(xhr.flash_message_html);`. So I think it is better to keep 03fa71a2e452c7ab68976d30c388d90c3ad1cade in this branch.

rmacklin: That's not necessary though. We can use an `id` or `js-` class on that element instead (in fact, that's what we should do anyway - to decouple our JS from our styles).

test/flow_tests/pictures_crud_test.rb
rmacklin: @shiboying replied here: https://github.com/AppFolioOnboarding/Boying_ImageUploader/commit/3258477709384e8d214dd9446aab5708d131123e#commitcomment-16222474 (where I replied again).

app/assets/javascripts/images/index/main.js
shiboying: Could you give me an example for it? I can just think of `share_modal` or `share_modal_manager` or `share_modal_monitor` but I don't think they are good fit.

app/assets/stylesheets/application.scss
```diff
@@ -3,3 +3,5 @@ $enable-flex: true;
 @import "image";
 @import "footer";
 @import "site";
+@import "align-horizontal";
+@import "tag-list";
```
aaronpanch: Not a big deal, but it's a nice convenience to alphabetize the imports (as long as there's no dependency issues-- but there shouldn't be)

app/assets/stylesheets/_tag-list.scss
```diff
@@ -0,0 +1,3 @@
+.tag-list {
+  flex: 1;
```
aaronpanch: I love how you're using flexbox! It makes layouts super easy :)  One thing: Flexbox hinges around the idea of a parent and child relationship.  To some degree, the parent defines the space available and the children get a portion of that space.  Currently your `tag-list` component asks for space in an assumed flex context; however, that means that this component isn't _self-contained_.  It requires the user to use another component to make this one work.  Looking at the usages below, I suggest you combine this component with `align-horizontal` (i.e. make `tag-list` a sub-component).  That way, the parent is clearly noted (with the `display: flex` property) and the child component can ask for a portion of that space. Let me know if you have questions :)

app/assets/javascripts/images/index/sharing_modal.js
rmacklin: I think you tried to address https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/51#discussion_r53520340, but the refactored code is not correct because I think you misunderstood what I was trying to say. Let's take a look at it tomorrow.

rmacklin:
I think you tried to address https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/51#discussion_r53520340, but the refactored code is not correct because I think you misunderstood what I was trying to say. Let's take a look at it tomorrow.

app/views/pictures/_share_modal.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="modal fade" id="shareModal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
```
rmacklin: `myModalLabel` should be `shareModalLabel` to match the `h4` on line 8.

app/views/pictures/_share_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= simple_form_for share_form,
+                    as: :share_form,
+                    remote: true,
+                    url: @picture.present? ? share_picture_path(@picture) : '', # url is required by simple form but we use data-path-template to set form action
+                    html: { 'data-path-template': share_picture_path('ID_PLACEHOLDER') } do |f| %>
+  <%= image_tag @picture.present? ? @picture.link : '', id: 'picture-preview', class: 'image' %>
```
rmacklin: This image tag is independent of the `form` and therefore does not need to be inside the `form` element, nor does it need to be inside this partial.

app/assets/javascripts/images/index/share_modal.js
```diff
@@ -0,0 +1,39 @@
+const NOT_FOUND = 404;
+const UNPROCESSABLE_ENTITY = 422;
+
+function shareModal(modalSelector, modalBodySelector, formSelector, pictureSelector) {
```
rmacklin: Hmm...I think we can get rid of some of these constructor arguments.

`modalBodySelector` is only used on line 15:
```js
$(modalBodySelector).html(xhr.responseJSON.form_html);
```
but this is equivalent to:
```js
$(formSelector).replaceWith(xhr.responseJSON.form_html);
```
so we can get rid of that.

Furthermore, we know there is only one `form` inside our share modal, so `formSelector` is not necessary because we can always get that element via `$shareModal.find('form')`.

Similarly, `pictureSelector` can be removed because we can get that element via `$shareModal.find('img')`.

app/assets/javascripts/images/index/share_modal.js
```diff
@@ -0,0 +1,39 @@
...
+  const $blankModal = $(modalSelector);
```
rmacklin: This isn't correct (and the application has a bug). As mentioned, I think I wasn't clear enough in https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/51#discussion_r53520340. Here's what I'm suggesting:
```js
const $shareModal = $(modalSelector);

$shareModal.on('ajax:success', (e, xhr, status, error) => { ... });

$shareModal.on('ajax:failure', (e, xhr, status, error) => { ... });

$shareModal.on('show.bs.modal', (event) => { ... });

$shareModal.on('hidden.bs.modal', () => { ... });
```
where we only do a single jQuery lookup, and everything else simply uses the cached result of that lookup (i.e. `$shareModal`). In order to make this work (and fix the current bug of not properly resetting the modal on hide), we need to store the blankModalHTML before the modal has been changed. We can do that here via:
```js
const blankModalHTML = $shareModal.html();
```
and then line 36 would be
```js
$shareModal.on('hidden.bs.modal', () => { $shareModal.html(blankModalHTML) });
```

app/assets/javascripts/images/index/share_modal.js
```diff
@@ -0,0 +1,39 @@
...
+
+  $(document).on('ajax:success', modalSelector, (e, xhr, status, error) => {
+    $(modalSelector).modal('hide');
+    $('.site__content > .container').prepend(xhr.flash_message_html);
```
rmacklin: We should probably use an `id` or a `js-` class for this.

app/assets/javascripts/images/index/main.js
rmacklin: Typically functions should be named as verbs describing what the method does (e.g. `add`, `calculateTaxes`, `renderForm`, etc.). The examples you just gave are nouns. Nouns are what we use when we name objects. In this case, I think we could actually refactor the function into a class, and then use one of the names you proposed (e.g. `ShareModal` or `ShareModalManager`).

rmacklin:
It's now hidden under "outdated diff", but I responded to your question about naming here: https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/51#discussion_r53746087.

app/assets/stylesheets/_align-horizontal.scss
```diff
@@ -0,0 +1,7 @@
+.align-horizontal {
+  display: flex;
+}
+
+.align-horizontal__tag-list {
+  flex: 1;
+}
```
rmacklin: I don't think this is what @aaronpanch was envisioning regarding the component name.

test/controllers/pictures_controller_test.rb
```diff
@@ -168,12 +168,59 @@ class PicturesControllerTest < ActionController::TestCase
...
+      xhr :post,
+          :share,
+          id: picture.id,
+          share_form: { recipient_email: 'invalid', message: 'hi' }
+    end
+
+    assert_response :unprocessable_entity
+
+    assert_includes json_response['form_html'], "id=\"new_share_form\" action=\"/pictures/#{picture.id}/share\""
+    assert_includes json_response['form_html'], "<img id=\"picture-preview\" class=\"image\" src=\"#{picture.link}\" alt=\"Img1\" />"
```
rmacklin: We can also assert that a validation error is showing.

app/assets/stylesheets/_align-horizontal.scss
shiboying: @rmacklin I think what Aaron means is `.tag-list` should stay in the same file and when using them `align-horizontal` should be parent and `tag-list` should be child. like:
```css
   <div class="align-horizontal">
     <div class="tag-list">
     </div>
   </div>
``` 
So considering the above two reasons, I think name it `.aligh-horizontal__tag-list` follows our naming convention.

@aaronpanch I think we discussed this file after yesterday `Frontend: Did you know` session, is my change correct?

aaronpanch: @shiboying Yes, your change follows the pattern and makes the parent-child relationship clearer :)  The last thing we should ensure is that the name of our components match what they represent.  In this case, calling our subcomponent `align-horizontal` doesn't say much about what it represents -- it could be tags or images -- and instead talks about what the component does.  From what I remember this has to do with the bottom part of the image card so a better alternative would talk about the the buttons or actions at the bottom of the card.

rmacklin: Addressed in https://github.com/AppFolioOnboarding/Boying_ImageUploader/commit/10da9a6772b4eeacb3d019f3650146609b67d910

app/views/pictures/_share_modal.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="modal fade" id="shareModal" tabindex="-1" role="dialog" aria-labelledby="shareModalLabel" aria-hidden="true">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+        <h4 class="modal-title" id="shareModalLabel">Email this picture</h4>
+      </div>
+      <div class="modal-body" id="shareModalBody">
```
rmacklin: The id `shareModalBody` is no longer necessary.

app/assets/javascripts/images/index/share_modal.js
```diff
@@ -0,0 +1,58 @@
...
+
+  handleSubmissionSuccess({flash_message_html}) {
+    this.$modal.modal('hide');
+    $('#site-content-container').prepend(flash_message_html);
+  }
+
+  handleSubmissionError({status, responseJSON}) {
+    switch (status) {
+      case UNPROCESSABLE_ENTITY:
+        this.$modal.find('form').html(responseJSON.form_html);
```
rmacklin: You should be using `replaceWith` not `html` here. (As mentioned in https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/51#discussion_r53745035)

app/assets/javascripts/images/index/share_modal.js
```diff
@@ -0,0 +1,58 @@
+import $ from 'jquery';
+
+const NOT_FOUND = 404;
+const UNPROCESSABLE_ENTITY = 422;
+
+class ShareModal {
+  constructor(modalSelector) {
+    this.$modal = $(modalSelector);
+    this.blankModalHTML = this.$modal.html();
```
rmacklin: :+1:

app/assets/javascripts/images/index/share_modal.js
```diff
@@ -0,0 +1,58 @@
...
+      this.handleSubmissionError(xhr);
+    });
+
+    this.$modal.on('show.bs.modal', (event) => {
+      this.setFormActionAndPicturePreview(event);
+    });
+
+    this.$modal.on('hidden.bs.modal', () => this.$modal.html(this.blankModalHTML));
+  }
+}
```
rmacklin: Awesome :+1:

This class is great!

app/assets/stylesheets/application.scss
```diff
@@ -1,5 +1,6 @@
 $enable-flex: true;
+@import "image-card-footer";
```
rmacklin: It's minor, but I think this list was intended to be alphabetized.

app/controllers/pictures_controller.rb
```diff
@@ -48,7 +74,14 @@ def picture_params
...
-    @picture = Picture.find_by_id(params[:id])
-    redirect_to root_path, flash: { danger: 'No such image!' } if @picture.blank?
+    @picture = Picture.find(params[:id])
+  end
+
+  def record_not_found
+    if request.xhr?
+      head :not_found
+    else
+      redirect_to root_path, flash: { danger: 'No such image!' } if @picture.blank?
```
rmacklin: Is this `if @picture.blank?` necessary?

app/views/pictures/index.html.erb
```diff
@@ -1,15 +1,17 @@
-<div class="container">
-  <% if @pictures.present? %>
+<% if @pictures.present? %>
+  <%= link_to('Upload New Image', new_picture_path, class: 'btn btn-info m-y-1') %>
+  <%= link_to('Show All Tags', tags_path, class: 'btn btn-info m-y-1') %>
+  <div id="picture_list" class="row">
+    <%= render @pictures %>
+    <%= render partial: 'share_modal' %>
```
rmacklin: I would move this outside of the `picture_list` element.

app/views/tags/index.html.erb
rmacklin: It would be better to move those to the other branch. Based on your comment here: https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/51#discussion_r53541831 and my response, it should no longer be necessary to have these changes on this branch now that our JS is using an `id`.

test/flow_tests/pictures_crud_test.rb
```diff
@@ -149,6 +149,55 @@ class PicturesCrudTest < FlowTestCase
...
+    control_labels = page.all('#new_share_form .control-label')
+    assert_equal '* Email', control_labels.first.text
+    assert_equal 'Message', control_labels.last.text
+
+    fill_in('* Email', with: 'iaminvalid')
+    fill_in('Message', with: email_message)
+
+    click_on('Send')
+    assert_equal 'is invalid', page.find('.has-error .help-block').text
```
rmacklin: We should verify that this error message is displayed for the email input.

test/flow_tests/pictures_crud_test.rb
```diff
@@ -149,6 +149,55 @@ class PicturesCrudTest < FlowTestCase
     assert_equal 'No such tag! ×', page.find('.alert').text
   end
 
+  test 'share pictures' do
+    picture = create_picture
+    valid_email = 'abc@gmail.com'
+    email_message = 'This email has a message.'
+
+    visit root_path
+    assert_not page.has_selector?('#new_share_form')
```
rmacklin: I'm actually surprised this line succeeds. The page actually has that selector even when the modal is not showing (the element is just hidden in that case). We should really be checking the visibility with:
```ruby
assert_not page.has_selector?('#new_share_form', visible: true)
```

test/flow_tests/pictures_crud_test.rb
```diff
@@ -149,6 +149,55 @@ class PicturesCrudTest < FlowTestCase
...
+
+    click_on('Share')
+    assert page.has_selector?('#new_share_form')
```
rmacklin: Again we should be checking the visibility:
```ruby
assert page.has_selector?('#new_share_form', visible: true)
```

test/flow_tests/pictures_crud_test.rb
```diff
@@ -149,6 +149,55 @@ class PicturesCrudTest < FlowTestCase
...
+
+    control_labels = page.all('#new_share_form .control-label')
+    assert_equal '* Email', control_labels.first.text
+    assert_equal 'Message', control_labels.last.text
```
rmacklin: Technically, this doesn't verify that there are no control labels between Email and Message. You could do that if you combined these assertions into:
```ruby
assert_equal ['* Email', 'Message'], control_labels.map(&:text)
```
which asserts that these are the _only_ two control labels.

test/flow_tests/pictures_crud_test.rb
```diff
@@ -149,6 +149,55 @@ class PicturesCrudTest < FlowTestCase
...
+
+    fill_in('* Email', with: 'iaminvalid')
+    fill_in('Message', with: email_message)
+
+    click_on('Send')
+    assert_equal 'is invalid', page.find('.has-error .help-block').text
+    assert page.has_selector?('#new_share_form')
```
rmacklin: Again we should be checking the visibility:
```ruby
assert page.has_selector?('#new_share_form', visible: true)
```

test/flow_tests/pictures_crud_test.rb
```diff
@@ -149,6 +149,55 @@ class PicturesCrudTest < FlowTestCase
...
+
+    fill_in('* Email', with: valid_email)
+    fill_in('Message', with: email_message)
+
+    click_on('Send')
+    assert_equal 'Share Successfully! ×', page.find('.alert').text
+    assert page.has_current_path?(root_path)
+
+    within_window open_new_window do
```
rmacklin: I think this limits your test to only be able to run using the drivers that implement `within_window`. That may be fine, but if you didn't want that limitation, you could replace lines 181-188 with `picture.destroy`. That has some of its own downsides because we are reaching into the low-level picture to call a method, so its a tradeoff (that way would also be faster than this way).

test/flow_tests/pictures_crud_test.rb
```diff
@@ -149,6 +149,55 @@ class PicturesCrudTest < FlowTestCase
...
+      visit root_path
+      visit picture_path(picture)
+      accept_alert do
+        click_on('Delete it')
+      end
+      assert picture_not_present?(picture.link)
+    end
+
+    click_on('Share')
+    assert page.has_selector?('#new_share_form')
```
rmacklin: Again we should be checking the visibility:
```ruby
assert page.has_selector?('#new_share_form', visible: true)
```

app/assets/javascripts/images/index/share_modal.js
shiboying: Oh, yes I forgot..

app/controllers/pictures_controller.rb
shiboying: No.Good catch

shiboying:
@rmacklin I have fixed my code based on your comments last night, thanks.

rmacklin:
You did not address this one: https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/51/files#r53890768, but I understand if you don't want to do that. If you have verified the acceptance criteria on your review app, you can merge this.

shiboying:
@rmacklin I will avoid this kind of mixing in the future. Thanks for your understanding. I have checked on my review app and it works properly, I will merge it.

PR #52, opened 2016-02-22
bs/dv - refactor the way to make footer fixed bottom and DRY containe…

shiboying:
…r into layout.


app/assets/stylesheets/_footer.scss
```diff
@@ -1,5 +1,14 @@
-.footer{
+.footer {
   background: #f7f7f9;
+  bottom: 0;
+  height: 180px;
+  position: absolute;
```
aaronpanch: I'm curious, why did you remove the flexbox sticky footer approach in favor of an absolute positioning.  They both work, but I'm curious about your rationale.  I'm asking because the absolute position method is slightly more brittle: we've hard-coded heights and widths to make it work versus using flexbox to do the computations automatically.

app/assets/stylesheets/_site.scss
```diff
@@ -1,10 +1,9 @@
...
-  flex-direction: column;
+html {
+  min-height: 100%;
+  position: relative;
 }
 
-.site__content {
-  flex: 1;
-  margin-top: 30px;
+.site-content {
```
aaronpanch: If this is the component defined in this file, it should be renamed to `_site-content.scss`. However, see comment above ^

app/assets/stylesheets/_site.scss
```diff
@@ -1,10 +1,9 @@
-.site {
-  display: flex;
-  min-height: 100vh;
-  flex-direction: column;
+html {
```
aaronpanch: We didn't go over this in the session: anytime we want to adjust styles for an element, whether that is the top-level `html`, `body`, or all links `a`, we should include them in a `base.scss` file since they describe a foundation or _base_ for components to build upon.

shiboying:
@aaronpanch The exact reason I cannot remember clearly but we did those changes was because we want to prepend a flash message to `.site-content > .container` and we extracted `.container` out from each template to application layout. 

But i changed it back to just use flex box in #51. Can you help me take a look at it?

And i have a question, what is different from name `.site__content` and `.site-content` and `.site > .site-content`? And also what is the different between add a sub class/or(component) inside and outside that file?

PR #53, opened 2016-02-23
Authentication

### LINKED ISSUE #48: Authentication
#### As an application hoster I want to add user authentication to my application.

- [ ] Users can sign up.
- [ ] Users can log in after they've signed up.
- [ ] Users can log out after logging in.
- [ ] Users can choose to be remembered after logging in.
- [ ] Multiple browsers can concurrently hold permanent cookies.
- [ ] Logged in users should not be able to register, nor login again.

Discussion topics:
- [ ] If we wanted to add more advanced features such as email activation and
  "forgot my password", should we continue to "roll our own" authentication?

Reference material:
- https://www.railstutorial.org/book/log_in_log_out

Further reading:
- https://github.com/plataformatec/devise

Dependencies:
- Capybara Testing

shiboying:
closes #48


shiboying:
@bboe I have modified my codes based on yesterday's pairing session, please help me review it. And just want to clarify, I didn't change to `resources :sessions` since I don't like it is shown as `sessions/new` in browser. So I just keep it the same way I did yesterday, for example, `get 'login' => 'sessions#new'` to make the path shown as `/login` in browser.

bboe:
> I didn't change to resources :sessions since I don't like it is shown as sessions/new in browser. So I just keep it the same way I did yesterday, for example, get 'login' => 'sessions#new' to make the path shown as /login in browser.

Thanks for clarifying. That's perfectly fine by me. Ignore my comments in the code.

Overall this is looking really great. There is some simplification that can be made, and a test or two that I think can be written. Great work!

Gemfile
```diff
@@ -26,6 +26,8 @@ gem 'acts-as-taggable-on', '~> 3.5'
 
 gem 'bootstrap', '~> 4.0.0.alpha3'
 
+gem 'bcrypt', '3.1.7'
```
rmacklin: Minor, but you could also have just uncommented line 33.

app/controllers/application_controller.rb
```diff
@@ -2,4 +2,5 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
+  include SessionsHelper
```
rmacklin: `SessionHelper` is intended to be used as the view helper for the `SessionController`'s views, and it's actually an antipattern to include that helper here in `ApplicationController`. I realize that this was actually suggested in the Rails Tutorial book, but I wanted to mention it.

A way you could work around this is to extract the methods that you need access to outside of the `SessionController` views into another module, and include that module here as well as in `SessionHelper`. A benefit of that is that then you'll _only_ have access to the methods you actually want available here, rather than all the methods in `SessionHelper`.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,33 @@
+class UsersController < ApplicationController
+  before_action :find_user, :only => [:show]
```
rmacklin: This is fine, but just wanted to mention that since it's only used for a single action, it's not gaining us too much as a `before_action`.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
```
rmacklin: Building on my last comment (https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/53#discussion_r54047443), if you had inlined the `before_action` into the `show` method, you could move the `not_found` handling into that action as well. Since that's the only action that tries to `find` a `User`, I think the code would actually be a bit easier to follow that way (because we would remove this level of indirection).

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def new
+    log_out if @user.blank?
```
rmacklin: I have the same question as Bryce: what else could we do if a logged in user tries to access the login page?

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def create
+    @user = User.new(users_params)
+
+    if @user.save
+      log_in @user
+      redirect_to @user, flash: { success: "Welcome #{@user.name}!" }
+    else
+      render 'new', status: :unprocessable_entity
```
rmacklin: Minor: you could use a symbol here.

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,37 @@
+module SessionsHelper
+  def current_user
+    if (user_id = session[:user_id])
+      @current_user ||= user_id && User.find_by(id: user_id)
```
rmacklin: What does `user_id &&` do here?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,24 @@
+class SessionsController < ApplicationController
+  def create
+    user = User.find_by(email: session_params[:email].downcase)
+    if user && user.authenticate(session_params[:password])
+      log_in user
+      remember(user) if session_params[:remember_me] == '1'
+      redirect_to user
+    else
+      flash.now[:danger] = 'Invalid email/password combination'
+      render 'new', status: :unprocessable_entity
```
rmacklin: Minor: you could use a symbol here.

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+    elsif (user_id = cookies.signed[:user_id])
+      user = User.find_by(id: user_id)
+      log_in user
```
rmacklin: Don't we need to check if the `user` is `authenticated?`? I just noticed your `User` model does not have that method. Did you follow https://www.railstutorial.org/book/log_in_log_out#code-authenticated_p?

app/models/user.rb
```diff
@@ -0,0 +1,27 @@
+class User < ActiveRecord::Base
+  attr_accessor :remember_token
+  VALID_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-.]+\.[a-z]+\z/i
```
rmacklin: This seems a bit different than the one you added for the share image story: https://github.com/AppFolioOnboarding/Boying_ImageUploader/blob/04d4ade4f85b4ac4b6da34d370fea78f63dc17f0/app/controllers/pictures_controller.rb#L8. We should consolidate into one "valid email" regex, and perhaps implement it as a custom validator which we can unit test once so we don't have to test different kinds of emails everywhere we validate emails.

app/models/user.rb
```diff
@@ -0,0 +1,27 @@
...
+  validates :name, presence: true, length: { maximum: 50 }
+  validates :email, presence: true, length: { maximum: 255 }, format: { with: VALID_EMAIL_REGEX }, uniqueness: { case_sensitive: false }
+
+  has_secure_password
+
+  validates :password, presence: true, length: { minimum: 6 }
+
+  def self.digest(string)
+    cost = ActiveModel::SecurePassword.min_cost ? BCrypt::Engine::MIN_COST :
+      BCrypt::Engine.cost
```
rmacklin: Minor: The indentation of this ternary is a bit odd. You could line up the two result operands like:
```ruby
cost = ActiveModel::SecurePassword.min_cost ? BCrypt::Engine::MIN_COST :
                                              BCrypt::Engine.cost
```
You could also have less leading whitespace with an `if`/`else`:
```ruby
cost = if ActiveModel::SecurePassword.min_cost
         BCrypt::Engine::MIN_COST
       else
         BCrypt::Engine.cost
       end
```

app/models/user.rb
```diff
@@ -0,0 +1,27 @@
...
+    BCrypt::Password.create(string, cost: cost)
+  end
+
+  def self.new_token
+    SecureRandom.urlsafe_base64
+  end
+
+  def remember
+    self.remember_token = User.new_token
```
rmacklin: Don't we also want to set a `remember_digest`? Did you follow https://www.railstutorial.org/book/log_in_log_out#sec-remember_token?

app/views/layouts/_navigation.html.erb
```diff
@@ -8,6 +8,18 @@
       <li class="nav-item">
         <%= link_to 'tags', tags_path, class: 'nav-link' %>
       </li>
+      <% if logged_in? %>
+        <li class="nav-item pull-xs-right">
+          <%= link_to 'Log out', logout_path, method: 'delete', class: 'nav-link' %>
+        </li>
+      <% else %>
+        <li class="nav-item pull-xs-right">
+          <%= link_to 'Signup', new_user_path, class: 'nav-link' %>
```
rmacklin: "Sign up" should be two words, like log in. We're using the verb form, rather than the noun "signup".

app/controllers/users_controller.rb
bboe: I concur with Richard.

app/views/sessions/new.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    </div>
+  </div>
+
+  <div class="row">
+    <div class="col-md-6 col-md-offset-3">
+      <%= simple_form_for :session, url: login_path do |f| %>
+        <%= f.input :email %>
+        <%= f.input :password %>
+        <%= f.input_field :remember_me, as: :boolean %>
+        <span>Remember me on this computer</span>
```
rmacklin: Can we use `f.input` and pass this as the `label` instead? It's a best practice to have `label`s for all `input`s (even checkboxes). It's good both for accessibility and for UX - clicking the label will toggle the checkbox.

app/views/sessions/new.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="container">
+  <div class="row">
+    <div class="m-t-1 m-b-1 col-md-6 col-md-offset-3">
+      <h3>Login</h3>
```
rmacklin: I think this should be the verb form: "Log in"

app/helpers/sessions_helper.rb
bboe: Boying had originally implemented the validation on permanent cookie. I asked her to remove it as I really don't like that it only allows a single browser to have a permanent cookie. We discussed that a better solution is to have an association list of cookie tokens, but that level of cookie validation goes well beyond the learning outcomes of this task in my opinion.

config/environments/production.rb
```diff
@@ -86,4 +86,6 @@
     domain: 'heroku.com',
     enable_starttls_auto: true
   }
+
+  config.force_ssl = true
```
rmacklin: You could have uncommented line 45 instead.

db/migrate/20160224183416_update_user_attributes_to_be_not_null.rb
```diff
@@ -0,0 +1,7 @@
+class UpdateUserAttributesToBeNotNull < ActiveRecord::Migration
+  def change
+    change_column_null :users, :name, false
+    change_column_null :users, :email, false
+    change_column_null :users, :password_digest, false
+  end
+end
```
rmacklin: You could combine the above four migrations into a single migration:
```ruby
class CreateUsers < ActiveRecord::Migration
  def change
    create_table :users do |t|
      t.string :name, null: false
      t.string :email, null: false
      t.string :password_digest, null: false

      t.timestamps null: false
    end

    add_index :users, :email, unique: true
  end
end
```
which would probably result in a faster migration process (and therefore less downtime for users of our application). For our little application, it won't make much of a difference, but I just wanted to mention it.

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,61 @@
...
+  setup do
+    @user = users(:tom)
+  end
+
+  test 'should get new' do
+    get :new
+    assert_response :success
+    assert_select '.simple_form' do |elements|
+      assert_equal [login_path], elements.map { |el| el[:action] }
+    end
```
rmacklin: I think you could do this more directly as `assert_select "form[action=#{login_path}]", 1`

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    post :create, session: { email: @user.email, password: 'password', remember_me: '0' }
+    assert_redirected_to user_path(@user)
+    assert is_logged_in?
+    assert_not is_permanent_logged_in?
+  end
+
+  test 'should log in with checking remember on this computer' do
+    post :create, session: { email: @user.email, password: 'password', remember_me: '1' }
+    assert_redirected_to user_path(@user)
+    assert is_logged_in?
```
rmacklin: This checks that _a user_ is logged in. It would be better to check specifically if _`@user`_ was logged in.

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    assert is_permanent_logged_in?
+  end
+
+  test 'should not log in with invalid input' do
+    post :create, session: { email: 'fake', password: 'faker' }
+    assert_response :unprocessable_entity
+    assert_equal 'Invalid email/password combination', flash[:danger]
```
rmacklin: We could also assert that the login form has been rerendered.

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    assert_not is_logged_in?
+    assert_not is_permanent_logged_in?
+  end
+
+  test 'should log out when remember on this computer is not checked' do
+    post :create, session: { email: @user.email, password: 'password', remember_me: '0' }
+    delete :destroy, id: @user.id
```
rmacklin: As Bryce mentioned, we can't hit multiple actions in a single controller test.

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    assert_not is_logged_in?
+    assert_redirected_to root_path
+  end
+
+  test 'should log out when remember on this computer is checked' do
+    post :create, session: { email: @user.email, password: 'password', remember_me: '1' }
+    delete :destroy, id: @user.id
```
rmacklin: As Bryce mentioned, we can't hit multiple actions in a single controller test.

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,37 @@
+require 'test_helper'
+
+class UsersControllerTest < ActionController::TestCase
+  test 'should access new' do
+    get :new
+    assert_response :success
```
rmacklin: What else can we assert here?

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
...
+    assert_equal 1, page.all('.has-error').count
+    assert_equal 'is too short (minimum is 6 characters)', page.find('.has-error .help-block').text
+    assert is_logged_out?
+
+    fill_in('* Password', with: 'abcdabcd')
+    fill_in('* Password confirmation', with: 'qwerqwer')
+    click_on('Create my account')
+
+    assert_equal 1, page.all('.has-error').count
+    assert_equal "doesn't match Password", page.find('.has-error .help-block').text
```
shiboying: @rmacklin I have noticed this should be ('.user_password_confirmation.has-error .help-block') and I have refactored it to a `error_message_for(attr)` private method in my local branch.

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  test 'can create a user' do
+    assert_difference('User.count', 1) do
+      post :create, user: { name: 'Alice', email: 'alice@example.com', password: 'valid_password', password_confirmation: 'valid_password'}
+    end
+
+    assert_redirected_to user_path(User.find_by(email: 'alice@example.com'))
```
rmacklin: What else can we assert here?

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  test 'cannot create a user with invalid input' do
+    assert_no_difference('Picture.count') do
+      post :create, user: { name: 'Alice', email: 'aliceexample', password: 'valid_password', password_confirmation: 'valid_password'}
+    end
+    assert_response :unprocessable_entity
```
rmacklin: What else can we assert here?

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  test 'can create a user' do
+    assert_difference('User.count', 1) do
+      post :create, user: { name: 'Alice', email: 'alice@example.com', password: 'valid_password', password_confirmation: 'valid_password'}
+    end
+
+    assert_redirected_to user_path(User.find_by(email: 'alice@example.com'))
+  end
+
+  test 'cannot create a user with invalid input' do
+    assert_no_difference('Picture.count') do
```
rmacklin: `Picture.count` :question:

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,37 @@
...
+      post :create, user: { name: 'Alice', email: 'aliceexample', password: 'valid_password', password_confirmation: 'valid_password'}
+    end
+    assert_response :unprocessable_entity
+  end
+
+  test 'can access show' do
+     user = User.create!(name: 'Alice', email: 'alice@example.com', password: 'valid_password', password_confirmation: 'valid_password')
+     get :show, id: user.id
+
+     assert_response :success
```
rmacklin: What else can we assert here?

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  test 'cannot show a non-existing user' do
+    get :show, id: 'nonexisting'
+
+    assert_redirected_to root_path
+    assert_equal "Sorry, this user doesn't exist", flash[:danger]
```
rmacklin: :+1:

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
+require 'flow_test_helper'
+
+class UsersCrudTest < FlowTestCase
+  setup do
+    Capybara.default_driver = Capybara.javascript_driver
+    @user = users(:tom)
```
rmacklin: This instance variable is only used in the second test. I would localize it to that test.

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
...
+  end
+
+  test 'sign up' do
+    visit new_user_path
```
rmacklin: Perhaps we can visit the `root_path` and then click the "Sign up" link to get here? That way we've verified that link works.

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
...
+      '* Name',
+      '* Email',
+      '* Password',
+      '* Password confirmation'
+    ]
+
+    assert_equal labels, page.all('.control-label').map(&:text)
+
+    click_on('Create my account')
+    assert_equal 3, page.all('.has-error').count
```
rmacklin: This is implicitly asserted by the assertion on line 31.

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
...
+    assert is_logged_out?
+
+    fill_in('* Name', with: 'Alice')
+    fill_in('* Email', with: 'alice@example.com')
+    fill_in('* Password', with: 'abcd')
+    fill_in('* Password confirmation', with: 'abcd')
+    click_on('Create my account')
+
+    assert_equal 1, page.all('.has-error').count
+    assert_equal 'is too short (minimum is 6 characters)', page.find('.has-error .help-block').text
```
rmacklin: We should verify this error is associated with the appropriate input.

test/flow_tests/users_crud_test.rb
rmacklin: We should verify this error is associated with the appropriate input.

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
...
+    assert_equal "doesn't match Password", page.find('.has-error .help-block').text
+    assert is_logged_out?
+
+    fill_in('* Password', with: 'abcdabcd')
+    fill_in('* Password confirmation', with: 'abcdabcd')
+    click_on('Create my account')
+
+    user = User.find_by_email('alice@example.com')
+    assert page.has_current_path?(user_path(user))
+    assert_equal "Welcome #{user.name}! ×", page.find('.alert').text
```
rmacklin: It would actually be a bit better to hardcode "Alice" here rather than `user.name`, because that way we have made sure the name we signed up with is the name that is displayed. (If our create action was silly and assigned every user's name to `Foo`, we wouldn't catch that with the way this test is currently written.)

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
...
+    assert is_logged_out?
+
+    fill_in('* Name', with: 'Bob')
+    fill_in('* Email', with: 'alice@example.com')
+    fill_in('* Password', with: 'abcdefgh')
+    fill_in('* Password confirmation', with: 'abcdefgh')
+    click_on('Create my account')
+
+    assert_equal 1, page.all('.has-error').count
+    assert_equal 'has already been taken', page.find('.has-error .help-block').text
```
rmacklin: We should verify this error is associated with the appropriate input.

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
...
+    assert is_logged_out?
+
+    fill_in('* Name', with: 'Alice')
```
rmacklin: Why change this to Alice?

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
...
+    fill_in('* Email', with: 'bob@example.com')
+    fill_in('* Password', with: 'abcdefgh')
+    fill_in('* Password confirmation', with: 'abcdefgh')
+    click_on('Create my account')
+
+    user = User.find_by_email('bob@example.com')
+    assert page.has_current_path?(user_path(user))
+    assert_equal "Welcome #{user.name}! ×", page.find('.alert').text
```
rmacklin: Same comment about hardcoding the name.

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
...
+    assert is_logged_in?
+  end
+
+  test 'log in and log out' do
+    visit login_path
```
rmacklin: Again, it would be better to get here from the root path to make sure our link works.

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
...
+    find('#login-btn').click
+
+    assert_equal 'Invalid email/password combination ×', page.find('.alert').text
+    assert is_logged_out?
+
+    fill_in('* Email', with: @user.email)
+    fill_in('* Password', with: 'password')
+    find('#login-btn').click
+
+    assert page.has_current_path?(user_path(@user))
```
rmacklin: What else could we assert about this page?

test/models/user_test.rb
```diff
@@ -0,0 +1,85 @@
+require 'test_helper'
+
+class UserTest < ActiveSupport::TestCase
+  def setup
+    @user = User.new(name: 'Bula', email: 'bula@example.com', password: 'fujiwater', password_confirmation: 'fujiwater')
+  end
+
+  test 'should be valid' do
+    assert @user.valid?
```
rmacklin: You could use `assert_predicate` and `refute_predicate` throughout this file for the better failure message.

test/flow_tests/users_crud_test.rb
rmacklin: Ah, I hadn't seen your comment before I posted mine. Great job noticing that :)

app/helpers/sessions_helper.rb
rmacklin: OK, sounds good.

test/flow_tests/users_crud_test.rb
shiboying: @rmacklin I want to test `two same name users with different email addresses can sign up`. I have a registered user: `name: Alice, email: alice@example.com` at line 56, here I want to test a user: `name: Alice, email:bob@example.com` can also be registered. 

FYI, line 65~72, test a user: `name: Bob, email:alice@example.com` cannot sign up since alice@example has already been taken.

So my test cases are:
```
1. name: Alice, email: alice@example.com
2. name: Bob, email: alice@example.com
3. name: Alice, email: bob@example.com
```

rmacklin: I see. The reason I asked is because ideally our flow tests should represent actions that real users would take. It's very unlikely that a user would try to sign up with one name, fail the email validation, and then change their name in addition to their email. A more realistic test case would be:

```
1. name: Alice, email: alice@example.com
2. name: Alice, email: alice@example.com
3. name: Alice, email: alice2@example.com
```

test/flow_tests/users_crud_test.rb
```diff
@@ -0,0 +1,154 @@
+require 'flow_test_helper'
+
+class UsersCrudTest < FlowTestCase
```
rmacklin: I wouldn't call this `UsersCrudTest` because it only tests the `C` and `R` from CRUD. Something like `RegistrationAndLoginTest` would be more accurate.

shiboying:
@bboe @rmacklin I have fixed my codes based on reviews. Please take a look at them, thanks!

app/helpers/login_helper.rb
```diff
@@ -0,0 +1,22 @@
+module LoginHelper
+  def current_user
+    if (user_id = session[:user_id])
+      @current_user ||= User.find_by(id: user_id)
+    elsif (user_id = cookies.signed[:user_id])
+      user = User.find_by(id: user_id)
+      # handle the case that cookies has user id stored
+      # but the user has been deleted in database
+      user.nil? ? forget : log_in(user)
```
rmacklin: `forget` is currently defined in `SessionsHelper`. We should move it here since we're using it in `LoginHelper`.

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,39 @@
+class SessionsController < ApplicationController
+  include SessionsHelper
```
rmacklin: Including this controller's view helper inside this controller is still an antipattern. But it's very easy to resolve this: we can just move the `forget` method to `LoginHelper`, and move the `log_out` and `remember` methods to be private methods in this controller.

app/helpers/login_helper.rb
```diff
@@ -0,0 +1,22 @@
+module LoginHelper
```
rmacklin: This isn't actually a view helper, so we could move it to the `controllers/concerns` directory.

app/controllers/pictures_controller.rb
```diff
@@ -5,9 +5,8 @@ class PicturesController < ApplicationController
   class ShareForm
     include ActiveModel::Model
     attr_accessor :recipient_email, :message
-    VALID_EMAIL = /\A[^@\s]+@([^@\s]+\.)+[^@\s]+\z/
 
-    validates :recipient_email, presence: true, format: { with: VALID_EMAIL }
+    validates :recipient_email, presence: true, email: true
```
rmacklin: Sweet :+1:

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,16 @@
+module SessionsHelper
```
rmacklin: As mentioned above, I think we'll be able to remove this file.

app/views/sessions/new.html.erb
```diff
@@ -0,0 +1,20 @@
...
+
+<div class="row">
+  <div class="col-md-6 col-md-offset-3">
+    <%= simple_form_for :session, url: login_path do |f| %>
+      <%= f.input :email %>
+      <%= f.input :password %>
+      <%= f.input :remember_me,
+                  as: :boolean,
+                  inline_label: true,
+                  label: 'Remember me on this computer' %>
```
rmacklin: :+1:

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  test 'should log in without checking remember on this computer' do
+    post :create,
+         session: {
+           email: @user.email,
+           password: 'password',
+           remember_me: '0'
+         }
+    assert_redirected_to user_path(@user)
+    assert is_logged_in?(@user.id)
```
rmacklin: I think it'd be a bit nicer if this method took a `User` instance instead of a user id.

test/flow_tests/registration_and_login_test.rb
```diff
@@ -0,0 +1,188 @@
...
+  setup do
+    Capybara.default_driver = Capybara.javascript_driver
+    Capybara.reset_sessions!
+  end
+
+  test 'sign up' do
+    visit root_path
+    click_on('Sign up')
+
+    assert is_logged_out?
```
rmacklin: We can also make an assertion about the current path.

test/flow_tests/registration_and_login_test.rb
```diff
@@ -0,0 +1,188 @@
...
+    assert_equal 'Welcome Alice! ×', page.find('.alert').text
+    assert is_logged_in?
+  end
+
+  test 'log in and log out' do
+    user = create_user
+
+    visit root_path
+    click_on('Log in')
+    assert is_logged_out?
```
rmacklin: We can also make an assertion about the current path.

test/flow_tests/registration_and_login_test.rb
```diff
@@ -0,0 +1,188 @@
...
+    assert_not page.has_link? 'Log out', login_path
+  end
+
+  def is_logged_in?
+    assert page.has_link? 'Log out', logout_path
+    assert_not page.has_link? 'Log in', login_path
+    assert_not page.has_link? 'Sign up', new_user_path
+  end
+
+  def render_user_show_page?(user_name, user_email)
```
rmacklin: I don't think `render` is the right word here. `on_user_show_page?` is probably a better name.

rmacklin:
@shiboying Awesome job addressing the review comments! I just left a few more.

bboe:
> Awesome job addressing the review comments! I just left a few more.

:+1: :+1:

app/controllers/application_controller.rb
```diff
@@ -2,4 +2,5 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
+  include LoginConcern
```
rmacklin: I think you only need this is the `SessionsController` and `UsersController`, so we should specifically include it there rather than here in `ApplicationController`.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,3 @@
 module ApplicationHelper
+  include LoginConcern
```
rmacklin: You could add a comment here indicating that this is including so that you'll have access to the `logged_in?` method.

shiboying: :+1:

app/controllers/application_controller.rb
shiboying: :+1:

